PR #239, opened 2024-05-21
[Feat]: Added ability to customize the pricing JSON

### LINKED ISSUE #240: [Feat]: Added ability to pass a custom URL to load the pricing JSON
### 🚀 What's the Problem?
As the default pricing JSON cannot cover Pricing for fine-tuned models, The cost is always 0 for fine tuned models or custom models which in not correct

### 💡 Your Dream Solution
Ability to pass a custom Pricing JSON URL which users can use to add custom model names that they use. This would also enable users to remove unused models from pricing JSON increasing evaluation speed.

### 🤔 Seen anything similar?
NA

### 🖼️ Pictures or Drawings
NA

pratiksinghchauhan:
#### Overview:
The PR adds the ability to use a custom JSON for pricing in case the user is using a fine tune or the model is not present in the default JSON

---

#### Changes:
- Add the ability to provide a custom JSON for pricing
- Update README

---

#### Checklist:
- [x] PR name follows conventional commits format: `[Feat]: ...` or `[Fix]: ....`

### DIFF
--- a/sdk/python/src/openlit/__helpers.py
+++ b/sdk/python/src/openlit/__helpers.py
@@ -142,7 +142,7 @@ def fetch_pricing_info(pricing_json=None):
                 f = open(pricing_json)
                 return json.load(f)
             except FileNotFoundError:
-                logger.error("File not found: %s", pricing_json)
+                logger.error("Pricing information file not found: %s", pricing_json)
             except json.JSONDecodeError:
                 logger.error("Error decoding JSON from file: %s", pricing_json)
             except Exception as file_err:
--- a/sdk/python/src/openlit/__helpers.py
+++ b/sdk/python/src/openlit/__helpers.py
@@ -2,8 +2,9 @@
 """
 This module has functions to calculate model costs based on tokens and to fetch pricing information.
 """
-
+import json
 import logging
+from urllib.parse import urlparse
 import requests
 import tiktoken
 from opentelemetry.trace import Status, StatusCode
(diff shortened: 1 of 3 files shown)


amanagarwal042:
Hey @pratiksinghchauhan for making the PR

One suggesstion - The current method would mean that the user would need to pass the complete json in the variable which seems a bit inconvenient

I was thinking [at this line](https://github.com/openlit/openlit/blob/f5b76f41aca2a085ba58718d97ca19f72c5e13e7/sdk/python/src/openlit/__helpers.py#L127) we could use the value of the `pricing_json` variable. If you feel this is unclear possible rename the variable to `pricing_json_url` ? and in cases when the url isnt passed, we use the default URL.

pratiksinghchauhan:
Sounds good. I will add support for both a file path and a URL. I would prefer it to be a file path that I can keep in the code where the SDK is used otherwise it is a bit of a hassle to host the JSON, but that would also be allowed.

pratiksinghchauhan:
@patcher9 I updated the PR as per the above conversation

sdk/python/src/openlit/__helpers.py
```diff
@@ -122,11 +123,35 @@ def get_audio_model_cost(model, pricing_info, prompt):
...
+    if pricing_json:
+        is_url = urlparse(pricing_json).scheme != ""
+        if is_url:
+            pricing_url = pricing_json
+        else:
+            try:
+                f = open(pricing_json)
+                return json.load(f)
+            except FileNotFoundError:
+                logger.error("File not found: %s", pricing_json)
```
amanagarwal042:
```suggestion
                logger.error("Pricing information file not found: %s", pricing_json)
```

amanagarwal042 (commented):
This is great, Small nitpick but overall I think its good to go

PR #266, opened 2024-06-05
[Feat]: adding playground to compare your evaluations from different providers

### LINKED ISSUE #132: [Feat]: Add Evaluation and Tuning Playground
### 🚀 What's the Problem?
Currently we cant experiment quickly and compare different models side-by-side on their response and related metrics like token usage, cost and latency

### 💡 Your Dream Solution
Add a playground where users can compare between different models from different providers side by siude on the same set of prompts.

### 🤔 Seen anything similar?
OpenAI Playground but think of it for multiple model providers and for testing a single prompt rather than the whole chat

### 🖼️ Pictures or Drawings
NA

AmanAgarwal041:
### Overview:
This PR introduces a feature to compare different providers for evaluation cost,  response time, different other parameters, response etc. Supported providers are listed below for now : 
- openai
- anthropic
- Mistral
- Cohere

Fixes #132

### Changes:
Adds a new route `/openground` to list all the previous comparisons
Adds a new route `/openground/new` to compare different providers side by side
Adds a new route `/openground/[id]` to display the previous comparison

This PR also addresses module upgrades.

### Visuals (If applicable):

Screenshot 2024-06-13 at 12 39 08 AM

Screenshot 2024-06-13 at 12 40 02 AM

Screenshot 2024-06-06 at 12 13 58 AM

Screenshot 2024-06-06 at 12 14 23 AM

Screenshot 2024-06-13 at 12 41 05 AM

### Checklist:


amanagarwal042:
Few comments and questions

- Instead of showing cost and tokens stuff first, Can we put response above as thats the first thing people want to compare before the dive into costs and stuff
- This will be placed below requests page right in side nav?
- If I have to make a new LLM compare request, I have to reload the page, can we have a CTA at the bottom and top right corner to allow people to make a new request
- In the table, cost and other field names are only on the left column, but not on the right, Can we add to both?
- Can we add units to the numbers like duration is in seconds so add `s` after the number and cost is USD so add `$` before the number

AmanAgarwal041:
> Few comments and questions
> 
> * Instead of showing cost and tokens stuff first, Can we put response above as thats the first thing people want to compare before the dive into costs and stuff
> * This will be placed below requests page right in side nav?
> * If I have to make a new LLM compare request, I have to reload the page, can we have a CTA at the bottom and top right corner to allow people to make a new request
> * In the table, cost and other field names are only on the left column, but not on the right, Can we add to both?
> * Can we add units to the numbers like duration is in seconds so add `s` after the number and cost is USD so add `$` before the number

All these suggestions are addressed.

AmanAgarwal041:
This PR also addresses module upgrades.

amanagarwal042:
Screenshot 2024-06-13 at 10 40 50 AM

I created a error request(I didnt enter the API keys) and now I can create a new request, Click on the button doesnt work


Even when I go from http://127.0.0.1:3000/openground and Click new, it goes back to the same screen

amanagarwal042 (approved):
LGTM

PR #538, opened 2024-12-03
[Fix]: Token usage when using chains in Langchain

Bassel-Elwakil:
### Overview:

- Added support for instrumenting Chain.invoke in LangChain.
- Support the instrumentation of LangChain's Chain.invoke method. Since the default LangChain chain response does not include response_metadata with token details, this implementation leverages the openai_tokens helper method from OpenLIT to calculate token usage for chains effectively.

### Visuals (If applicable):
image

### Checklist:

## Summary by Sourcery

Add support for token usage instrumentation in LangChain's Chain.invoke method by using the openai_tokens helper method to calculate token usage when response_metadata is missing. Update the version to 1.32.8 in pyproject.toml.

Bug Fixes:
- Fix token usage calculation in LangChain's Chain.invoke method by leveraging openai_tokens helper method when response_metadata is missing.

Enhancements:
- Enhance Chain.invoke method in LangChain to support token usage instrumentation even when response_metadata is not available.


sdk/python/src/openlit/instrumentation/langchain/langchain.py
```diff
@@ -507,12 +507,26 @@ def wrapper(wrapped, instance, args, kwargs):
...
+                prompt = ""
+                if hasattr(response, 'response_metadata') and response.response_metadata:
+                    token_usage = response.response_metadata.get("token_usage", {})
+                    input_tokens = token_usage.get("prompt_tokens", 0)
+                    output_tokens = token_usage.get("completion_tokens", 0)
+                    model = response.response_metadata.get("model_name", "gpt-4")
+                    prompt = "" if isinstance(args[0], list) else args[0]
+                else:
+                    if not isinstance(response, dict) or "output_text" not in response:
+                        logger.warning(f"Ignoring uninstrumented response: {response}")
```
amanagarwal042: Dont think we need to log this as we generally also dont

amanagarwal042 (changes requested):
Bit unsure of this but also since I havent tested this (I think I am just missing context on why so just some more info for me might help).

Could you share the sample snippet to this with? Also you might wanna fix the pylint changes to get the workflow to pass

Bassel-Elwakil:
@patcher9 

The default LangChain Chain.invoke response often omits response_metadata, which includes crucial details like token usage (prompt_tokens, completion_tokens) and the model name.
To address this, used the openai_tokens helper method from OpenLIT to manually calculate token usage when response_metadata is absent.
This ensures token-level observability for all LangChain chains, enhancing OpenLIT's ability to track metrics like token counts, model usage, and costs during LangChain workflows.

How It Works:
When response_metadata is present in the chain's response, token details are extracted directly.
If response_metadata is missing (which is common for default chains), the implementation calculates token usage using openai_tokens, leveraging the page_content of input documents and the output_text of the chain.
The calculated metrics (e.g., input_tokens, output_tokens, model) are recorded in spans and metrics using OpenLIT.

I attached a screenshot showing that OpenLIT instrumentation logs token usage for this operation, even without response_metadata for langchain chains.

sdk/python/src/openlit/instrumentation/langchain/langchain.py
```diff
@@ -549,7 +562,7 @@ def wrapper(wrapped, instance, args, kwargs):
                                     input_tokens + output_tokens)
                 span.set_attribute(SemanticConvetion.GEN_AI_USAGE_COST,
                                     cost)
-                if trace_content:
+                if trace_content and hasattr(response, 'content'):
```
amanagarwal042: I think if response doesnt have content attribute then even the prompt will not be added to the span events, So either lets make the second check a sub conditional or by default log "" for response

amanagarwal042: Also I think you should use general_tokens rather than openai_tokens as people can also tend to use anthropic/other LLMs with langchain and general_tokens would be better

amanagarwal042 (changes requested):

amanagarwal042 (approved):

PR #557, opened 2025-01-02
[Fix]: Install prisma at buildtime instead of at runtime

dbirks:
### Overview:

Currently when the docker image starts up, it does a `npm install -g prisma` each time. Normally I wouldn't notice, but in my company all https traffic needs to be encrypted with an internal TLS cert, making it error out when I run it locally on my laptop:

```
Error: request to https://binaries.prisma.sh/all_commits/11f085a2012c0f4778414c8db2651556ee0ef959/linux-musl-openssl-3.0.x/schema-engine.gz.sha256 failed, reason: self-signed certificate in certificate chain
```

Moving the install step from running at runtime in the `entrypoint.sh`, to running at buildtime in the `Dockerfile`.

Thanks for working on openlit! I've been liking it

### Checklist:

## Summary by Sourcery

Build:
- Install Prisma in the Dockerfile during the build process.


amanagarwal042:
Thanks @dbirks for the PR!

@AmanAgarwal041 can you take a pass at this?

amanagarwal042:
Apologies for the delay in review of this one @dbirks but looks good and really appreciate the change!

amanagarwal042 (approved):

PR #718, opened 2025-05-09
[Fix]:  Keep the original messages order while showing the prompt

Gwenn-LR:
### Overview:

Hi, first of all thanks to all the team and the community for your amazing work ! I've opened this PR since I've noticed that the list of messages sent to my client was not ordered as expected. After checking my code, I've finally found the origin of my issue : while `openlit` creare a span, all messages are extracted and formatted but as a dictionnary which is not ordered.

So if my objective is for the trace to be a close representation of the request to my client, I have 2 issues :
- A source list of X message with alternating roles would be formatted as a dictionnary of 5 entries (one for each of the elements among `['user', 'assistant', 'system', 'tool', 'developer']`.
- An ordered list of message would than be concatenated among the content of the unordered dictionnary (even if it seems to unpack its values as a FIFO, which currently follow the sequence `['user', 'assistant', 'system', 'tool', 'developer']`.

That's why I've used a list, to keep the hierarchy, while keeping the idea to concatenate message if able. This change also have required to adapt the concatenation method, to handle a list of dictionnaries as input.

I don't know if this behaviour is known by your team and expected, or if my changes would break something else, but if it can help, here is my little contribution to your helpful project ! :)

### Visuals (If applicable):
Here if my input list of messages:
{15992631-E5AB-46EE-8030-3AD2D6435C73}

Here if the formatted output dictionnary:
{2A8707EF-9BDC-4A7C-A830-BAD0149B4919}

After the changes:
{7DBA9B8B-2514-49C8-AF48-5CC423C060A6}

### Checklist:

## Summary by Sourcery

Refactor message processing to store messages in a list, preserving their original order.

Bug Fixes:
- Correct message ordering in prompts, previously affected by role-based grouping and loss of sequence for multiple messages of the same role interspersed with other roles.
- Ensure message content is accurately concatenated for consecutive messages of the same role within the preserved order.


amanagarwal042:
Hey @Gwenn-LR Some CI issues in the pipeline you might see. But before fixing I would say we do plan to remove events logs for prompts and responses and continue using span attributes as thats what the broader community is looking to adopt aswell. So this change might not render say in the next 1 month

Gwenn-LR:
Hi @patcher9, thank you for your answer. No problem, I understand focusing your work on the approach the community is adopting ! I don't think I understand clearly the implications of your "plan to remove events logs for prompts and responses and continue using span attributes", but if you need help with the `Ollama` instrumentation, I'll be glad to contribute  ! :)

Meanwhile, thank you for your time and have a nice day !

PR #769, opened 2025-06-26
[Fix]: Handle KeyError for OpenAI tool calls with no content

### LINKED ISSUE #767: [Bug]: KeyError: 'content' in async OpenAI wrapper during tool calls
### 🐛 What's Going Wrong?
The OpenLIT async wrapper for the OpenAI client (async_openai.py) raises a KeyError: 'content' when it intercepts a chat completion request where one of the messages is an assistant message containing tool_calls.

According to the OpenAI API specification, when an assistant message includes tool_calls, the content field can be null or entirely absent. The current OpenLIT instrumentation does not handle this valid case and fails when it tries to directly access the content key, causing the trace to be marked as an error.

### 🕵️ Steps to Reproduce
This issue can be reproduced with any standard tool-calling sequence. The error occurs on the second API call, which includes the history of the assistant's decision to call a tool.

A minimal example:

```
import asyncio
from openai import AsyncOpenAI
# Assuming 'openlit' is configured to instrument the client

client = AsyncOpenAI()

async def main():
    # This message history simulates the context for a follow-up call 
    # after a tool has been called and returned a result.
    # The second message is the one that causes the error.
    messages_with_tool_call_history = [
        {"role": "user", "content": "What is the weather in Berlin?"},
        # This assistant message has 'tool_calls' but no 'content' key.
        {"role": "assistant", "tool_calls": [
            {"id": "call_123", "type": "function", "function": {"name": "get_weather", "arguments": '{"location": "Berlin"}'}}
        ]},
        # The result of the tool call.
        {"role": "tool", "tool_call_id": "call_123", "name": "get_weather", "content": "The weather is sunny."}
    ]

    try:
        # This API call will be intercepted by OpenLIT's wrapper,
        # which will then raise the KeyError while parsing the messages list.
        await client.chat.completions.create(
            model="gpt-4o-mini",
            messages=messages_with_tool_call_history
        )
        print("API call succeeded without instrumentation error.")
    except KeyError as e:
        print(f"OpenLIT instrumentation failed with KeyError: {e}")

asyncio.run(main())
```

### 🎯 What Did You Expect?
I expected the OpenLIT wrapper to successfully process all valid OpenAI API message formats, including assistant messages that contain tool_calls but no content. The trace should be recorded correctly without raising an exception.

### 📸 Any Screenshots?
The following exception is captured by our observability platform, pointing directly to the issue in the OpenLIT wrapper:

```
{
  "name": "exception",
  "timestamp": "2025-06-26T11:43:39.817760Z",
  "attributes": {
    "exception.type": "KeyError",
    "exception.message": "'content'",
    "exception.stacktrace": "Traceback (most recent call last):\n  File \".../openlit/instrumentation/openai/async_openai.py\", line 736, in wrapper\n    content = message[\"content\"]\nKeyError: 'content'\n",
    "exception.escaped": "False"
  }
}
```

### 💻 Your Setup
OpenLIT Version: Langfuse with OpenLit Intstrumentation
OpenLIT SDK Version: openlit v1.34.
Deployment Method: local
Relevant Libraries: openai, autogen-agentchat v0.6.1

### 📝 Additional Notes
The recommended fix is to make the message parsing more robust by using the .get() method for dictionary access.

In openlit/instrumentation/openai/async_openai.py:

Change this line:
`content = message["content"]`

to this

`content = message.get("content", "")`

This ensures that the wrapper does not fail on valid API payloads and correctly handles tool-calling scenarios.

ChrisBlaa:
### Overview:
Closes #767

### Checklist:

## Summary by Sourcery

Bug Fixes:
- Default to an empty string when 'content' key is absent in OpenAI tool calls

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "openlit"
-version = "1.34.5"
+version = "1.34.6"
 description = "OpenTelemetry-native Auto instrumentation library for monitoring LLM Applications and GPUs, facilitating the integration of observability into your GenAI-driven projects"
 authors = ["OpenLIT"]
 license = "Apache-2.0"
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "openlit"
-version = "1.34.6"
+version = "1.34.7"
 description = "OpenTelemetry-native Auto instrumentation library for monitoring LLM Applications and GPUs, facilitating the integration of observability into your GenAI-driven projects"
 authors = ["OpenLIT"]
 license = "Apache-2.0"
--- a/sdk/python/src/openlit/instrumentation/openai/async_openai.py
+++ b/sdk/python/src/openlit/instrumentation/openai/async_openai.py
@@ -733,7 +733,7 @@ def async_chat_completions(version, environment, application_name,
                     formatted_messages = []
                     for message in message_prompt:
                         role = message["role"]
-                        content = message["content"]
+                        content = message.get("content", "")
 
                         if isinstance(content, list):
                             content_str = ", ".join(


amanagarwal042 (approved):

amanagarwal042:
Thanks @ChrisBlaa , Will release it in a few mins

Issue #997, opened 2026-02-10
Bug: Async Generator Syntax Error Preventing Initialization


ialisaleh:
### Component

OpenLIT Python SDK

### What happened?

**Description**
OpenLIT fails to initialize in the latest version (v1.36.8) with the error:

```bash
------------------------------ Captured log call -------------------------------
ERROR    openlit:__init__.py:445 Error during openLIT initialization: 'return' with value in async generator (async_agno.py, line 783)
```

**Root Cause**
I started facing this issue 3 days ago, when this PR [#959](https://github.com/openlit/openlit/pull/959) was merged, because the error occurs in `sdk/python/src/openlit/instrumentation/agno/async_agno.py` at line 783, which contains changes to an async generator function.

### Steps to reproduce

Install the following dependencies:
```bash
$ pip install "openlit==1.36.8" opentelemetry-sdk openai
```

**Case 1: Following Official Documentation (Simple Init)**

**Code:**
```python
import openlit
from openai import OpenAI

openlit.init()
client = OpenAI(api_key="sk-test-key")

response = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "Hello"}]
)
print(response)
```

**Error Output:**
`ERROR    openlit:__init__.py:445 Error during openLIT initialization: 'return' with value in async generator (async_agno.py, line 783)`

**Result:**
No spans are created, and instrumentation doesn't work.

**Case 2: Using Custom Tracer (OpenTelemetry Integration)**

**Code:**
```python
import openlit
from openai import OpenAI
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor, ConsoleSpanExporter

tracer_provider = TracerProvider()
tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))
tracer = tracer_provider.get_tracer(__name__)

openlit.init(
    otel_tracer=tracer,
    otlp_endpoint=None,
)
client = OpenAI(api_key="sk-test-key")

response = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "Hello"}]
)
print(response)
```

**Error Output:**
`ERROR    openlit:__init__.py:445 Error during openLIT initialization: 'return' with value in async generator (async_agno.py, line 783)`

**Result:**
No spans are created, and instrumentation doesn't work.

### Expected behavior

The instrumentation should work fine in the latest version.

[...]

PR #998, opened 2026-02-11
fix: Async Generator Syntax Error Preventing Initialization

ialisaleh:
**Issue number**: #997 

### Change description:

This PR fixes a critical initialization bug in the v1.36.8 release. The `async_workflow_run_wrap` function is an async generator, but was using `return await result` on line 783, which violates Python's async generator syntax rules (PEP 525). This was changed to `yield await result` to fix the error `'return' with value in async generator` that prevented OpenLIT from initializing. This one-word fix restores full functionality.

### Checklist

* [ ] Changes are documented

### Acknowledgment

## Summary by Sourcery

Bug Fixes:
- Correct async generator behavior in async_agno instrumentation by yielding awaited results instead of returning them, preventing initialization errors.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/agno/async_agno.py
+++ b/sdk/python/src/openlit/instrumentation/agno/async_agno.py
@@ -780,7 +780,7 @@ def async_workflow_run_wrap(
                 async for event in result:
                     yield event
             else:
-                return await result
+                yield await result
             return
 
         workflow_name = getattr(instance, "name", "unknown_workflow")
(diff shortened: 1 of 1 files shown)


ialisaleh:
Hi @amanagarwal042 - This PR resolves the initialization bug and works correctly with the documented usage examples (tested locally). However, the CI tests are failing because they were written expecting the buggy behavior.

The `async_workflow_run_wrap` function is designed to always return an async generator (as indicated by the "maintain generator contract" comment). My change makes line 783 consistent with this design by using `yield` instead of `return`.

**9 tests need updates:**
test_assemblyai.py, test_chroma.py, test_elevenlabs.py, test_groq.py (2 tests), test_mistral.py (3 tests), test_transformers.py

Pattern: Change `result = await func()` to `async for result in func(): pass`

Happy to update the tests if needed, or the team can handle it during review. Thanks!

amanagarwal042:
@ialisaleh I think this was fixed in #999 . Mind confirming or adapting your PR for the change?

The PR is not yet released so can wait for you

PR #1117, opened 2026-04-14
fix: correct payload field names when saving a custom model

### LINKED ISSUE #1107: Bug: Cannot add new model in Manage Models page
### Component

OpenLIT

### What happened?

When adding a new model, the openlit Docker container generates the following log:

POST /api/openground/models - Body: {

  "provider": "google",

  "model": {

    "model_id": "gemini-3-flash-preview",

    "displayName": "Gemini 3 Flash Preview",

    "contextWindow": 4096,

    "inputPricePerMToken": 0,

    "outputPricePerMToken": 0,

    "capabilities": []

  }

}

CustomId type: undefined value: undefined

Validation failed - provider: google model.id: undefined model.displayName: Gemini 3 Flash Preview

There seems to be a mistmatch between the body payload model.model_id attribute and the model.id attribute expected by the API backend.

That issue has been there since 1.17.0 (when that page appeared)

This is problematic if we want to add estimated pricing for "custom" or new models (such as gemini-3-flash-preview or the newer OpenAI/Anthropic models)

### Steps to reproduce

1. Open the OpenLIT dashboard
2. Navigate to the Manage Model page (http://localhost:3000/settings/manage-models)
3. Enter a new model and try to save

### Expected behavior

It should allow the user to enter a new model with a not-yet-used model ID

### Environment

_No response_

### Additional context

POST /api/openground/models - Body: {

  "provider": "google",

  "model": {

    "model_id": "gemini-3-flash-preview",

    "displayName": "Gemini 3 Flash Preview",

    "contextWindow": 4096,

    "inputPricePerMToken": 0,

    "outputPricePerMToken": 0,

    "capabilities": []

  }

}

CustomId type: undefined value: undefined

Validation failed - provider: google model.id: undefined model.displayName: Gemini 3 Flash Preview

### Pre-submission checklist

Yes, I am willing to submit a PR!

geomarceau:
fixes #1107

The POST request from the Manage Models page was sending the wrong field names, causing saves to always fail with a 400.

Two mismatches: the model identifier was sent as model.model_id but the API reads model.id, and the UUID for updates was sent as id at the top level but the API expects customId.

Because model.id was always undefined, the validation check would reject every request. The customId mismatch also meant updates were never recognized as updates and fell through to inserts instead.

Fixed both field names in the handleSave payload in model-editor-panel.tsx.

### DIFF
--- a/src/client/src/components/(playground)/openground/model-editor-panel.tsx
+++ b/src/client/src/components/(playground)/openground/model-editor-panel.tsx
@@ -21,7 +21,7 @@ import { toast } from "sonner";
 
 interface CustomModel extends ModelMetadata {
   id: string; // UUID from database
-  model_id: string; // Model identifier like "gpt-4o"
+  modelId: string; // Model identifier like "gpt-4o"
 }
 
 interface ModelEditorPanelProps {
@@ -50,7 +50,7 @@ export default function ModelEditorPanel({
   const [showDeleteDialog, setShowDeleteDialog] = useState(false);
   const [formData, setFormData] = useState<Partial<CustomModel>>({
     id: "", // UUID (only for existing custom models)
-    model_id: "", // Model identifier
+    modelId: "", // Model identifier
     displayName: "",
     contextWindow: 4096,
     inputPricePerMToken: 0,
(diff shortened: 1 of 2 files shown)


AmanAgarwal041:
Thanks @geomarceau for opening this pull request. 
We are moving the implementation of models. So the issue won't be valid and would be fixed with https://github.com/openlit/openlit/pull/1117.
Hence closing this PR.

PR #1134, opened 2026-04-23
fix(ts/langchain): propagate AIMessage.tool_calls to gen_ai.output.messages

imxinde:
## Summary

The TypeScript LangChain instrumentation drops `AIMessage.tool_calls` entirely. When the model emits a tool call, the span records `finish_reason: "tool_calls"` but `gen_ai.output.messages` only contains a text part — making the tool invocation invisible in the trace.

## The bug

In `sdk/typescript/src/instrumentation/langchain/wrapper.ts` the handler only reads `gen.text` / `msg.content` from the generations array:

```ts
// Content
const content = gen?.text || msg?.content || '';
if (content && typeof content === 'string' && ...) {
    completionContent = content;
}
// ↓ msg.tool_calls is never looked at
```

Then it builds the output messages without the `toolCalls` argument:

```ts
span.setAttribute(
  SemanticConvention.GEN_AI_OUTPUT_MESSAGES,
  OpenLitHelper.buildOutputMessages(completionContent, finishReason)
  //                                                    ^ no toolCalls
);
```

`helpers.buildOutputMessages(text, finishReason, toolCalls)` already supports the third parameter — the `openai`, `anthropic`, and `groq` wrappers all pass it. LangChain is the only mainstream wrapper that doesn't.

## Repro

A LangGraph ReAct agent calling a tool (verified with `@langchain/langgraph` 1.x + `@langchain/openai` 1.x, `openlit` 1.11.0):

**Before:**
```json
{
  "parts": [{"type":"text","content":"I'll help you get Shanghai's temperature..."}]
}
```

**After:**
```json
{
  "parts": [
    {"type":"tool_call","id":"call_1","name":"get_weather","arguments":{"city":"Shanghai"}},
    {"type":"tool_call","id":"call_2","name":"calculator","arguments":{"expression":"22*9/5+32"}}
  ]
}
```

## Changes

1. Extract `tool_calls` from `AIMessage` / `additional_kwargs` during generation iteration.
2. Normalise LangChain's `{ id, name, args }` shape → `{ id, type, name, arguments }` expected by `buildOutputMessages`.
3. Pass the collected `toolCalls` as the third argument to `buildOutputMessages`.
4. Emit `gen_ai.output.messages` even when `completionContent` is empty but there ARE tool_calls (e.g. pure tool-use turns).
5. Set the flat `gen_ai.tool.{name,call.id,call.arguments}` attributes that the OpenAI wrapper already emits — useful for backends that filter on span tags without parsing nested JSON.
6. Export `OpenLITCallbackHandler` so it can be unit-tested directly.

## Tests

New `langchain-wrapper.test.ts` with 4 regression tests:
- standard LangChain `AIMessage.tool_calls` (`{ id, name, args }`) → tool_calls reach `buildOutputMessages`
- text-only completion → `toolCalls` param is `undefined` (unchanged behaviour)
- tool-only completion (empty content) → `gen_ai.output.messages` still gets set
- legacy shape via `additional_kwargs.tool_calls` → still picked up

All 25 test suites / 141 tests pass (24/137 before + 1 suite / 4 tests).

## Test plan
- [x] `npm test` in `sdk/typescript/` — all pass
- [x] `npm run build` — clean
- [x] `npx eslint` on modified files — no new lint errors vs baseline
- [x] Verified against a live LangGraph ReAct agent reporting to an OTLP backend

## Summary by Sourcery

Propagate LangChain AIMessage.tool_calls into recorded gen_ai.output.messages and surface tool-call metadata on spans while adding regression tests.

New Features:
- Include LangChain tool_calls in gen_ai.output.messages and expose flattened tool metadata span attributes for easier backend filtering.

Bug Fixes:
- Fix loss of LangChain AIMessage.tool_calls in TypeScript instrumentation so tool invocations are visible in traces, including tool-only turns and legacy additional_kwargs.tool_calls.

Enhancements:
- Export OpenLITCallbackHandler from the LangChain wrapper to enable direct unit testing of its behavior.

Tests:
- Add langchain-wrapper regression tests covering tool_calls propagation, text-only completions, tool-only completions, and legacy additional_kwargs.tool_calls handling.

### DIFF
--- a/sdk/typescript/src/instrumentation/__tests__/langchain-wrapper.test.ts
+++ b/sdk/typescript/src/instrumentation/__tests__/langchain-wrapper.test.ts
@@ -234,4 +234,99 @@ describe('LangChain wrapper — tool_calls propagation', () => {
       { id: 'legacy_1', type: 'function', name: 'lookup', arguments: '{"k":"v"}' },
     ]);
   });
+
+  it('does not merge tool_calls across multiple generations (n > 1)', async () => {
+    // When the provider returns multiple choices, LangChain still collapses to
+    // one assistant message and `finishReason`/`completionContent` take the
+    // last writer. Tool calls must follow the same last-writer-wins rule —
+    // otherwise calls from different choices get flattened into one message
+    // and flat tool-name / id attributes become cross-choice joins.
+    const span = makeSpan();
+    seedRun(handler, 'run-tc-5', span);
+
+    const output = {
+      generations: [
+        [
+          {
+            text: '',
+            message: {
+              content: '',
+              tool_calls: [{ id: 'choice_a_1', name: 'tool_a', args: {} }],
+              response_metadata: { finish_reason: 'tool_calls' },
+            },
+          },
+          {
+            text: '',
+            message: {
+              content: '',
+              tool_calls: [{ id: 'choice_b_1', name: 'tool_b', args: {} }],
+              response_metadata: { finish_reason: 'tool_calls' },
+            },
+          },
+        ],
+      ],
+    };
+
+    handler.handleLLMEnd(output, 'run-tc-5');
+    await new Promise((r) => setImmediate(r));
+
+    const [, , toolCalls] = (OpenLitHelper.buildOutputMessages as jest.Mock).mock.calls[0];
+    // Expect only the last generation's tool_calls — NOT both merged.
+    expect(toolCalls).toHaveLength(1);
+    expect(toolCalls[0].name).toBe('tool_b');
+
+    // Flat attributes must not be a cross-choice join like "tool_a, tool_b".
+    const setAttr = span.setAttribute as jest.Mock;
+    expect(setAttr).toHaveBeenCalledWith(SemanticConvention.GEN_AI_TOOL_NAME, 'tool_b');
+    expect(setAttr).toHaveBeenCalledWith(SemanticConvention.GEN_AI_TOOL_CALL_ID, 'choice_b_1');
+  });
+
+  it('survives unserialisable tool arguments (circular refs / BigInt) and still ends the span', async () => {
+    // If JSON.stringify throws, the outer try/catch in _finalizeLLMSpan would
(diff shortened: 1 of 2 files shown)


imxinde:
Thanks for the reviews! Pushed [a7c05ea](https://github.com/openlit/openlit/commit/a7c05eab) addressing the two Copilot findings:

**1. Tool calls across `n > 1` generations** — previously `toolCalls` was accumulated across all entries in `output.generations` and then collapsed into one assistant message, so multiple choices would get merged (and the flat `gen_ai.tool.name` / `gen_ai.tool.call.id` attributes became cross-choice joins). Now the accumulator resets on each new generation that carries `tool_calls`, which lines up with the existing last-writer-wins semantics of `finishReason` / `completionContent`.

**2. Unsafe `JSON.stringify` on tool arguments** — circular refs / BigInt would throw inside `_finalizeLLMSpan`, the outer `try/catch` would swallow it, and the span would be left un-ended with the run entry leaked in the `spans` map. Each per-tool-call serialisation is now individually guarded and falls back to `'[unserializable]'` so `span.end()` and `spans.delete()` always run.

Added two regression tests that both fail against the un-fixed version:
- `does not merge tool_calls across multiple generations (n > 1)`
- `survives unserialisable tool arguments (circular refs / BigInt) and still ends the span`

**Re: @sourcery-ai comment on mock path** — verified the mock path is correct. The test mocks `'../../../src/helpers'` (resolves to `sdk/typescript/src/helpers.ts`) and `wrapper.ts` imports `'../../helpers'` (also resolves to `sdk/typescript/src/helpers.ts`); Jest matches mocks against the resolved module path, not the specifier string. I also ran a sentinel probe (mocked `buildOutputMessages` to return `'MOCK_SENTINEL_VALUE'`) and confirmed the value flowed through `wrapper.ts` to `span.setAttribute(gen_ai.output.messages, ...)` — the mock is actually wired in.

Full suite: 143/143 pass (up from 141 before these follow-ups), build clean, no new lint errors.

amanagarwal042 (approved):

PR #1267, opened 2026-06-10
Fix: decrypt vault value in OpenGround config path

### LINKED ISSUE #1266: Fix: vault values returned undecrypted in getOpenGroundConfigWithSecret — all OpenGround calls fail with bad auth
## Bug

`getOpenGroundConfigWithSecret` in `src/client/src/lib/platform/providers/config.ts` returns the raw vault value without decrypting it. This means:

- **Anthropic**: provider receives `invalid x-api-key` error
- **OpenAI-compatible providers**: receive the literal `enc:v1:...` ciphertext as the API key

Reproducible on any docker-compose install once a Vault secret is created.

## Root cause

`upsertSecret` in `src/client/src/lib/platform/vault/index.ts` encrypts on write, but `getOpenGroundConfigWithSecret` reads `vaultRecord.value` raw without calling `decryptValue`.

## Fix

One-line change in `config.ts`:

```diff
import Sanitizer from "@/utils/sanitizer";
+import { decryptValue } from "@/utils/crypto";
 ...
     return {
       data: {
         ...config,
-        apiKey: vaultRecord.value,
+        apiKey: decryptValue(vaultRecord.value),
         vaultKey: vaultRecord.key,
       },
     };
```

`decryptValue` short-circuits on non-`enc:v1:` values so it is safe for both encrypted and plaintext rows.

SelfParody:
Fixes #1266

## Summary

`getOpenGroundConfigWithSecret` was returning the raw encrypted vault value (`enc:v1:...`) as the API key, causing all OpenGround provider calls to fail with auth errors.

### Change

```diff
+import { decryptValue } from "@/utils/crypto";
 ...
-        apiKey: vaultRecord.value,
+        apiKey: decryptValue(vaultRecord.value),
```

`decryptValue` short-circuits on non-`enc:v1:` prefixed values so this is safe for both encrypted and legacy plaintext rows.

## Summary by Sourcery

Bug Fixes:
- Fix OpenGround provider authentication failures by decrypting encrypted vault API key values before use.

### DIFF
--- a/src/client/src/lib/platform/providers/config.ts
+++ b/src/client/src/lib/platform/providers/config.ts
@@ -1,6 +1,7 @@
 import { dataCollector } from "@/lib/platform/common";
 import getMessage from "@/constants/messages";
 import Sanitizer from "@/utils/sanitizer";
+import { decryptValue } from "@/utils/crypto";
 import { OPENLIT_VAULT_TABLE_NAME } from "@/lib/platform/vault/table-details";
 import { OPENLIT_PROVIDERS_TABLE_NAME } from "./table-details";
 
@@ -122,7 +123,7 @@ export async function getOpenGroundConfigWithSecret(
 		return {
 			data: {
 				...config,
-				apiKey: vaultRecord.value,
+				apiKey: decryptValue(vaultRecord.value),
 				vaultKey: vaultRecord.key,
 			},
 		};


AmanAgarwal041 (approved):

AmanAgarwal041:
@SelfParody Thanks for the contribution 🚀

PR #1272, opened 2026-06-12
fix: respect externally set NEXTAUTH_URL in entrypoint.sh

### LINKED ISSUE #1080: Bug: Next Auth Url always is localhost:$PORT
### Component

OpenLIT

### What happened?

When deploying openlit, the entrypoint.sh always add NEXTAUTH_URL as a localhost bases environment to /etc/environment file and evaluate it with `. /etc/environment`, so the previous environment variable added to the deployment is replaced.

### Steps to reproduce

1. set NEXTAUTH_URL to anything else
2. run scripts/entrypoint.sh
3. try login with oauth application

### Expected behavior

It should be use the value on NEXTAUTH_URL of environment and not localhost:$PORT (like :3000)

### Environment

_No response_

### Additional context

_No response_

### Pre-submission checklist

None

Oxygen56:
## Summary
- Added guard to only set NEXTAUTH_URL default when not already provided
- Allows users to configure a custom NEXTAUTH_URL for production deployments

Fixes #1080

## Test Plan
- [x] Shell syntax validated with bash -n
- [x] NEXTAUTH_URL env var is respected when set externally

## Summary by Sourcery

Bug Fixes:
- Respect externally provided NEXTAUTH_URL values by avoiding overwriting them in entrypoint.sh.


amanagarwal042 (approved):
Thanks @Oxygen56 as always!

Issue #1245, opened 2026-06-03
Feature: add PostgreSQL (pg) instrumentation parity with Python psycopg in TS SDK


amanagarwal042:
### Component

OpenLIT

### What feature would you like to see?

Add auto-instrumentation for the Node `pg` (and/or `postgres`) driver, matching what Python's `psycopg` instrumentation captures for GenAI / pgvector workloads.

- New TS instrumentation directory: `sdk/typescript/src/instrumentation/pg/` (or `psycopg/` for naming parity)
- Capture queries, including pgvector similarity operations
- Honor `captureDbParameters` toggle (parity with Python's `capture_db_parameters`)

### Why do you need this?

Python has `sdk/python/src/openlit/instrumentation/psycopg/`. TS RAG users on Postgres/pgvector get no traces today. Note: `@opentelemetry/instrumentation-pg` exists as a community option, but Python's psycopg instrumentation adds GenAI-specific attributes (vector queries, db parameter capture toggle) that the generic community one doesn't.

### Additional context

- Python reference: `sdk/python/src/openlit/instrumentation/psycopg/`
- Could either: (a) wrap `pg` directly, or (b) extend the community `@opentelemetry/instrumentation-pg` with OpenLIT-specific GenAI attributes (similar to how OTel community HTTP instrumentations are loaded today in `sdk/typescript/src/instrumentation/index.ts`).

### Pre-submission checklist

- [x] I searched existing issues and didn't find a duplicate

### Are you willing to submit PR?

None

chippleh1392:
@amanagarwal042 I can take this one. My read is that pg users get no traces today since there's only the vector-DB instrumentations, no Postgres one. I'd follow the same pattern as chroma/qdrant, name it `pg` after the driver, and use the Python psycopg module as the reference for the attributes, including the pgvector similarity operators, with a mocked trace-comparison test like the others. I'll get a PR up.

icohangar-ops:
I'd like to work on this. Planning to add a `pg` instrumentation under `sdk/typescript/src/instrumentation/pg/` that wraps `Client.prototype.query` and `Pool.prototype.query` (handling callback, promise, and async forms), mirroring the Python `psycopg` instrumentor: SQL operation/table parsing, pgvector similarity-operator detection (`<=>`, `<->`, `<#>`), a `captureDbParameters` toggle for query parameters, and the same `db.*` semantic-convention attributes/metrics. Will open a PR shortly.

PR #1312, opened 2026-06-30
feat(typescript): add PostgreSQL (pg) instrumentation (#1245)

icohangar-ops:
> [!IMPORTANT]
> Issue-first: this PR implements [#1245](https://github.com/openlit/openlit/issues/1245).

**Issue number**: #1245

### Change description

Adds auto-instrumentation for the Node `pg` driver in the **TypeScript SDK**, bringing it to parity with the Python SDK's `psycopg` instrumentation.

New directory `sdk/typescript/src/instrumentation/pg/` (`index.ts` + `wrapper.ts`), registered in `instrumentation/index.ts`, with `'pg'` added to the `InstrumentationType` union and the PostgreSQL semantic-convention constants added to `semantic-convention.ts` (mirroring the Python attribute names/values).

**What it captures (parity with Python psycopg):**
- Wraps `Client.prototype.query` directly (and `Pool.prototype.query` when it is a distinct implementation — `pg`'s `Pool.query` delegates to `Client.query`, so we avoid double-counting). Handles all three `query()` call shapes:
  - **callback**: `client.query(text, values?, cb)` — wraps the callback, finalizes the span on `(err, res)`.
  - **promise**: `await client.query(text, values?)` — finalizes on resolve/reject.
  - **submittable** (`Query`/`Cursor`): listens for `end`/`error` events.
- Parses the SQL operation verb and target table → span name `"<OP> <table>"`, `db.operation.name`, `db.collection.name`.
- Detects **pgvector** similarity operators and sets `db.search.similarity_metric`: `<=>` → `cosine`, `<->` → `l2`, `<#>` → `inner_product` (matches Python `detect_special_features`), and reflects the metric in `db.query.summary`.
- **`captureDbParameters`** config toggle (parity with Python's `capture_db_parameters`) — records query parameters per-index as `db.query.parameter.<n>` (and per-key for object params), with the same sanitization (`NULL`, `<bytes:N>`, length truncation at 256). **Off by default** since parameters may contain PII/secrets.
- Sets the same `db.*` attributes as Python: `db.system.name=postgresql`, `db.namespace` (database), `db.query.text` (gated on `captureMessageContent`), `db.query.summary`, `db.response.returned_rows` (from `result.rowCount`), `db.client.operation.duration`, `server.address`/`server.port` (defaults `localhost:5432`).

**Tests** (`pg-trace-comparison.test.ts`, mirrors the Chroma test, mocks `pg` — no live DB): basic query span, pgvector query, and the `captureDbParameters` toggle on/off, plus callback/promise/error paths. 30 tests, all green. Full TS suite: 503/503 passing; `lint` and `build` clean.

**Deviations from Python psycopg:** the Node `pg` driver has a single `query()` entry point, so this wraps that rather than the psycopg `Cursor.execute/executemany/copy/callproc` + `Connection.commit/rollback` + pool `getconn` surface. The operation verb, pgvector detection, parameter capture, and attribute set are all preserved; `db.batch.size` and commit/rollback/COPY/CALL semconv constants are added for completeness/parity.

`dist/` is intentionally left untracked (not tracked upstream).

### Checklist
* [x] PR name follows conventional commit format: `feat: ...`
* [x] I have reviewed the contributing guidelines
* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?
* [x] Changes are documented

### Acknowledgment

## Summary by Sourcery

Add PostgreSQL (pg) auto-instrumentation to the TypeScript SDK to align database tracing semantics with the existing Python psycopg integration.

New Features:
- Introduce a dedicated pg instrumentation module that wraps the Node pg driver to produce OpenTelemetry client spans for PostgreSQL queries, including callback, promise, and submittable forms.
- Support configuration to optionally capture query parameters for pg spans, mirroring the Python SDK capture_db_parameters behavior.

Enhancements:
- Extend semantic-convention constants with PostgreSQL-specific database attributes, SQL operation names, pgvector similarity metrics, and batch size metadata for cross-language parity.
- Register pg as a first-class instrumentation type and hook it into the central instrumentation registry for automatic setup.

Tests:
- Add comprehensive pg instrumentation tests that validate span attributes, pgvector similarity handling, parameter capture behavior, and error paths against the Python psycopg reference implementation.


icohangar-ops:
Thanks @sourcery-ai for the review — all three points are addressed in 4230925:

1. **SpanStatusCode** — replaced the hard-coded `span.setStatus({ code: 1 })` with `span.setStatus({ code: SpanStatusCode.OK })`, importing `SpanStatusCode` from `@opentelemetry/api`. That was the only hard-coded status numeric (error status is set via `OpenLitHelper.handleException`, which uses the enum).

2. **Double-instrumentation** — fixed by wrapping **only `Client.prototype.query`** and removing the `Pool.prototype.query` wrapping entirely. In `pg`, `Pool.query()` checks out a pooled Client and delegates to that Client's `query()`, so client-level wrapping already captures pool-routed queries; wrapping both produced two spans (Pool + child Client) per logical pool query. This matches the OTel community `pg` instrumentation. Added a test asserting a `pool.query()` that delegates to `client.query()` yields exactly **one** span, plus an instrumentation-level test confirming `Client.prototype.query` is wrapped while `Pool.prototype.query` is left untouched.

3. **Submittable/Cursor coverage** — added tests for the `result.on('end' | 'error')` branch: (a) a fake submittable that emits `end` finalizes the span (attributes set + `end()` called once), (b) one that emits `error` calls `OpenLitHelper.handleException` and ends the span, and (c) a guard test ensuring the span is finalized only once if `end` fires after `error`.

`lint` + `build` clean; full TS suite green (37 suites / 509 tests, up from 503; the pg suite is now 36 tests).

Issue #1319, opened 2026-07-01
Bug: Disable metrics not working


guilhermetakataext-cmd:
### Component

OpenLIT Python SDK

### What happened?

The parameter disable_metrics is not working, even when setting it to True, my meter provider is still trying to export metrics

### Steps to reproduce

openlit.init(disable_metrics=True_

### Expected behavior

It should not create and try to export metrics

### Pre-submission checklist

- [x] I searched existing issues and didn't find a duplicate

VanshikaMehta18:
Hi, I would like to take this one! 

I'll trace where disable_metrics is read in openlit.init() and check whether it gates meter provider initialization or just sets a flag that isn't checked before the provider starts. Will add a test confirming no meter provider is created when disable_metrics=True.

Could you please assign this to me?

amanagarwal042:
@VanshikaMehta18 assigned this to you now

PR #1329, opened 2026-07-03
fix(python-sdk): skip meter provider setup when disable_metrics=True

VanshikaMehta18:
openlit.init() always called setup_meter(), creating OTLP exporters even when metrics were disabled. Gate setup behind disable_metrics to match the TypeScript/Go SDK behavior.

> [!IMPORTANT]  
> 2. PR name follows conventional commit format: `feat: ...` or `fix: ....`

**Issue number**:

### Checklist

* [ ] PR name follows conventional commit format: `feat: ...` or `fix: ....`
* [ ] I have reviewed the [contributing guidelines](https://github.com/openlit/openlit/blob/main/CONTRIBUTING.md)
* [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/openlit/openlit/pulls) for the same update/change?
* [ ] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented

### Acknowledgment


AmirF194 (approved):
Consolidated my notes into the review below, please disregard this duplicate.

AmirF194 (approved):
Nice, focused fix. I checked this out and traced it through the SDK, and it holds up.

What I verified:
- Root cause matches #1319: `init()` called `setup_meter()` unconditionally, so `disable_metrics=True` still built the meter provider and OTLP exporters. Gating the call behind `if not disable_metrics` is the right fix and matches the TS/Go SDK behavior.
- The seams are clean. `metrics_dict` and `err` are initialized to `None` up front, so the later `if err:` and the `update_config(..., metrics_dict, ...)` call are safe when metrics are off. Downstream readers are all guarded too: `log_agent_invocation`/`log_agent_tool_error` check `if metrics:`, and every instrumentor that records metrics (openai, langgraph, crewai, crawl4ai) gates on `if not disable_metrics and metrics:`, so a `None` dict never gets indexed. The GPU and system-metrics blocks were already behind `not disable_metrics`.
- The regression test is real. I ran it in a clean Python 3.12 container both ways: it passes on your branch, and when I revert just the guard hunk it fails with `metrics_dict` populated (the exact #1319 symptom) while the `disable_metrics=False` case keeps passing. That is what you want a regression test to do.

Two optional, non-blocking notes:
- The PR body leaves "Issue number" blank. Since this resolves #1319 and the repo is issue-first, adding `Fixes #1319` to the description would auto-close the issue on merge.
- `test_disable_metrics_false_sets_up_meter` creates a real MeterProvider but never shuts it down, so at interpreter exit the periodic exporter thread prints `ValueError: I/O operation on closed file`. Harmless and unrelated to your change, but a `teardown_method` that shuts the provider down would keep the output clean.

One more heads-up, unrelated to your code: the Python SDK test job is currently `python-tests.yml.hold`, so this regression test will not actually run in CI (only pylint and leak-check run on PRs here). Worth re-enabling that workflow, or running the test locally before merge, so this stays covered.

Neither blocks. LGTM, leaving the merge call to the maintainers.

amanagarwal042 (approved):
sweet thanks!

PR #1342, opened 2026-07-07
fix: capture OpenAI reasoning content

### LINKED ISSUE #1341: OpenAI/OpenRouter instrumentation drops reasoning_content from generation spans
## Summary

OpenLit's OpenAI instrumentation records normal chat completion output content, reasoning token counts, and request metadata, but it does not record returned reasoning text on the generation span. This affects OpenRouter models that expose reasoning via OpenAI-compatible Chat Completions fields such as `delta.reasoning_content` / `delta.reasoning`.

The semantic convention already exists in OpenLit:

```python
SemanticConvention.GEN_AI_CONTENT_REASONING == "gen_ai.content.reasoning"
```

Azure AI Inference and Cohere instrumentation already use this attribute, so the OpenAI instrumentation looks like an oversight rather than a deliberate omission.

## Reproduction shape

Use an OpenAI-compatible client against OpenRouter with a reasoning-capable model and streaming enabled:

```python
client.chat.completions.create(
    model="anthropic/claude-sonnet-4.6",
    messages=[{"role": "user", "content": "Reason step by step, then answer."}],
    stream=True,
    extra_body={"reasoning": {"effort": "medium"}},
)
```

Streaming chunks include reasoning text in the delta:

```json
{"choices":[{"delta":{"reasoning_content":"Let me work through this..."}}]}
```

OpenLit's `openai/utils.py` currently appends only `delta.content` to `scope._llmresponse`, so the reasoning text never reaches the span.

## Expected

When `capture_message_content=True`, the OpenAI instrumentation should set:

```text
gen_ai.content.reasoning = "<aggregated reasoning text>"
```

This should work for both streaming and non-streaming chat completions.

## Actual

No `gen_ai.content.reasoning` attribute is set by OpenAI instrumentation.

## Existing in-repo precedent

Azure AI Inference already implements the same behavior:

- streaming accumulation: `sdk/python/src/openlit/instrumentation/azure_ai_inference/utils.py`
- canonical span attr: `SemanticConvention.GEN_AI_CONTENT_REASONING`
- non-streaming extraction from `choices[0].message.reasoning_content`

Cohere also sets `GEN_AI_CONTENT_REASONING` for tool planning content.

## Suggested fix

Port the Azure pattern into `sdk/python/src/openlit/instrumentation/openai/utils.py`:

- accumulate `delta.reasoning_content` or `delta.reasoning` in `process_chat_chunk`
- extract `message.reasoning_content` or `message.reasoning` in `process_chat_response`
- set `SemanticConvention.GEN_AI_CONTENT_REASONING` in `common_chat_logic` inside the existing `capture_message_content` block

I can send a PR with tests if this approach looks acceptable.

josefonte:
## What

Captures returned reasoning text in the OpenAI instrumentation and stores it on the generation span as `gen_ai.content.reasoning`.

## Why

OpenAI-compatible providers such as OpenRouter can return reasoning text in chat completion payloads via `reasoning_content` or `reasoning`. The current OpenAI instrumentation only accumulates `delta.content`, so reasoning text is dropped even when `capture_message_content=True`.

OpenLit already has precedent for this attribute in Azure AI Inference and Cohere instrumentation.

Fixes #1341.

## How

- accumulate `delta.reasoning_content` / `delta.reasoning` during streaming chat completions
- extract `message.reasoning_content` / `message.reasoning` for non-streaming chat completions
- set `SemanticConvention.GEN_AI_CONTENT_REASONING` inside the existing content-capture gate
- add unit tests for streaming, non-streaming, and content-capture-disabled behavior

## Tests

```bash
poetry run pytest tests/test_openai_error_attributes.py tests/test_openai_reasoning_content.py -q
```

## Summary by Sourcery

Capture reasoning content returned by OpenAI-compatible chat completions and expose it as span attributes when message content capture is enabled.

New Features:
- Record reasoning text from streaming and non-streaming OpenAI-compatible chat responses onto spans via the gen_ai.content.reasoning semantic attribute.

Bug Fixes:
- Ensure reasoning content from OpenAI-compatible providers is no longer dropped when capture_message_content is enabled.

Tests:
- Add unit tests covering streaming and non-streaming reasoning content capture and behavior when content capture is disabled.

Chores:
- Reset Openlit configuration to defaults in OpenAI instrumentation tests to avoid cross-test interference.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/openai/utils.py
+++ b/sdk/python/src/openlit/instrumentation/openai/utils.py
@@ -96,6 +96,14 @@ def handle_not_given(value, default=None):
     return value
 
 
+def extract_reasoning_content(payload):
+    """Return OpenAI-compatible reasoning text from a message or delta payload."""
+    if not isinstance(payload, dict):
+        return ""
+    reasoning = payload.get("reasoning_content") or payload.get("reasoning")
+    return reasoning if isinstance(reasoning, str) else ""
+
+
 def format_content(messages):
     """
     Format the messages into a string for span events.
@@ -676,6 +684,12 @@ def process_chat_chunk(scope, chunk):
         content = delta.get("content")
         if content:
             scope._llmresponse += content
+
+        reasoning_content = extract_reasoning_content(delta)
+        if reasoning_content:
+            if not hasattr(scope, "_reasoning_content"):
+                scope._reasoning_content = ""
+            scope._reasoning_content += reasoning_content
 
         # Handle tool calls in streaming - optimized
         delta_tools = delta.get("tool_calls")
@@ -1583,6 +1597,11 @@ def common_chat_logic(
     # capture is enabled.
     if capture_message_content:
         _set_span_messages_as_array(scope._span, input_msgs, output_msgs)
+        if hasattr(scope, "_reasoning_content") and scope._reasoning_content:
+            scope._span.set_attribute(
+                SemanticConvention.GEN_AI_CONTENT_REASONING,
+                scope._reasoning_content,
+            )
         if system_instr:
             scope._span.set_attribute(
                 SemanticConvention.GEN_AI_SYSTEM_INSTRUCTIONS,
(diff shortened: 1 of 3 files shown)


amanagarwal042 (approved):
This is great @josefonte !

PR #1350, opened 2026-07-08
docs: add TokenLab OpenAI-compatible monitoring example

hedging8563:
## Summary
- Show OpenLIT instrumentation with the OpenAI SDK pointed at TokenLab.
- Keep the change docs-only and scoped to the project's existing OpenAI-compatible configuration surface.

## Validation
- git diff --check

## Summary by Sourcery

Documentation:
- Add a Python example showing OpenLIT instrumentation of the OpenAI SDK configured to send requests to TokenLab via a custom base URL.


hedging8563:
Thanks for the review. Addressed in 31b0635 by clarifying that claude-sonnet-5 is a TokenLab-supported model ID that can be replaced with any model from TokenLab's /v1/models endpoint, and that TOKENLAB_API_KEY should be kept in environment variables or a secret manager rather than hard-coded.

PR #1368, opened 2026-07-14
fix: reject disabled offline evaluation types

Sakshamm-Goyal:
## Summary

Reject explicitly requested offline evaluation types when they are disabled in the tenant configuration.

## Root cause

`runOfflineEvaluation` selected any requested configured type without checking its `enabled` flag, so a caller could run a scorer that the dashboard had disabled.

## Scope

- Return a clear error for requested disabled evaluation types before model execution.
- Preserve the existing default fallback only when no types are explicitly requested.
- Add a regression test proving disabled scorers do not call the evaluator.

## Validation

- `npm test -- --runInBand src/__tests__/lib/platform/evaluation/evaluation.test.ts`
- `npm run lint`

`npm run typecheck` remains blocked by two pre-existing implicit-any errors in `src/components/(playground)/manage-dashboard/board-creator/components/code-editor.tsx`.

## Summary by Sourcery

Ensure offline evaluations respect tenant configuration for enabled evaluation types.

Bug Fixes:
- Prevent runOfflineEvaluation from executing explicitly requested evaluation types that are disabled in the tenant configuration.

Tests:
- Add a regression test verifying that disabled evaluation types are rejected with a clear error and do not invoke the evaluator.


amanagarwal042:
@Sakshamm-Goyal closing this as this was already merged in main. Looking forward to future PRs ad issues from you. Thanks!

Issue #1369, opened 2026-07-14
[Bug]: Anthropic streaming instrumentation - `ERROR opentelemetry.context Failed to detach context` / `ValueError: Token ... was created in a different Context`


ron-42:
## OpenLIT version

`1.40.3` (PyPI), confirmed same code present on `main` at time of filing.

## Environment

- Python 3.12
- `anthropic` SDK (async client, `client.messages.stream(...)`)
- `opentelemetry-api`/`opentelemetry-sdk` (contextvars-based runtime context)
- App is an async FastAPI service calling Anthropic through
  [agno](https://github.com/agno-agi/agno) (agno itself is not instrumented
  by OpenLIT here, plain `httpx`/`anthropic` async calls under the hood)

## Description

When an Anthropic streaming response (`messages.stream()` /
`beta.messages.stream()`) is instrumented by OpenLIT and the underlying
async generator is not driven to exhaustion in the same asyncio Task that
started it, we intermittently see this in logs:

```
ERROR    opentelemetry.context | Failed to detach context
Traceback (most recent call last):
  File ".../opentelemetry/context/__init__.py", line 155, in detach
    _RUNTIME_CONTEXT.detach(token)
  File ".../opentelemetry/context/contextvars_context.py", line 53, in detach
    self._current_context.reset(token)
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x...> at 0x...> was created in a different Context
```

`opentelemetry.context.detach()` already catches this internally (it's a `logger.exception` inside a `except Exception:` block), so it does not crash the request, but it pollutes logs at ERROR level and indicates the span
context lifecycle in the Anthropic instrumentation is not safe against a generator being finalized outside its originating Task.

## Root cause

`sdk/python/src/openlit/instrumentation/anthropic/async_anthropic.py` attaches an OTel context `Token` when a stream is opened and detaches it when the stream wrapper is closed:

- `wrapper()`, streaming branch: `context_api.attach(ctx)` / `context_api.detach(token)` around L132-143.
- `TracedMessageStreamManager`: `self._token = context_api.attach(ctx)`  (~L356) on entry, `context_api.detach(self._token)` (~L375) on exit.

[...]

Issue #861, opened 2025-08-30
[Bug]: Langchain: ERROR: opentelemetry.context Failed to detach context with langgraph


ianchi:
### 🐛 What's Going Wrong?
Running a langgraph graph with `astream` inside an async fastapi app.
The langchain instrumentor is throwing lots of these errors, and in the end no spans are generated in the trace

```text
ERROR:  opentelemetry.context Failed to detach context
Traceback (most recent call last):
  File ".venv/lib/python3.11/site-packages/opentelemetry/context/__init__.py", line 155, in detach
    _RUNTIME_CONTEXT.detach(token)
  File ".venv/lib/python3.11/site-packages/opentelemetry/context/contextvars_context.py", line 53, in detach
    self._current_context.reset(token)
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x72e67a44f560> at 0x72e63d24cc40> was created in a different Context
```
They all come from this line:

instrumentation.langchain.callback_handler.py 225 in _end_span

called from `on_chain_end` and   `on_llm_end`

### 💻 Your Setup
- OpenLIT Version: no backend, using opentelemetry collector
- OpenLIT SDK Version: 1.35.1
- Deployment Method: none

### 👐 Want to Help Make It Happen?
- [x] Yes, I'd like to volunteer and help out with this!

amanagarwal042:
Hey @ianchi , To confirm, You are interested in raising a PR for the fix? If yes I could help guide you to the correct files that might need to be updated

ianchi:
Hi @patcher9, yes, I could open a PR.
I did some quick checks, and know the files involved (langchain instrumentation), but I couldn't find what makes the contextvar get mixed.
I think it might have something to do with the fact that there are three frameworks using contextvars and switching contexts:
- fastapi
- langgraph
- opentelemetry 

I could not find a way to debug the contextvar stack to get any grasp of what is hapening.
If you have some ideas I could try them and test it.

ianchi:
continuing on this topic, I saw the same problem in Azure Langchain's instrumentation and in OpnLLMetery.

Take a look at this part of the [code](https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py#L208) specially the comments.

It is a hackish way of ignoring the error buts avoid polluting the log.

How do you find that solution?


Regarding the spans not being generated, I belive it is not related to the contextvar errors.
[...]

PR #1370, opened 2026-07-14
fix: detect Task mismatch before OTel context detach in Anthropic streaming

ron-42:
**Issue number**: #1369

### Change description:

Fixes `ValueError: <Token var=<ContextVar name='current_context' ...> at 0x...> was created in a different Context` / `Failed to detach context` (ERROR-level, logged by `opentelemetry/context/__init__.py`) in OpenLIT's Anthropic instrumentation.

**Root cause**: `TracedAsyncMessageStreamManager`/`TracedMessageStreamManager` (and the streaming branch of `wrapper()`) in `anthropic.py`/`async_anthropic.py` call `context_api.attach()` on stream entry and `context_api.detach()` on stream exit. If a caller abandons a streaming response mid-iteration (e.g. breaks out of an `async for` early, or an outer async generator wrapping the stream is never driven to `StopAsyncIteration`), Python's asyncgen GC finalizer hook later drives `aclose()` on the abandoned generator — and does so in a *fresh* asyncio Task, not the one that originally called `attach()`. `contextvars.Token.reset()` requires the exact `Context` it was created in, so detaching from the wrong Task always raises `ValueError`.

**Fix**: `safe_detach()` (`openlit/__helpers.py`) now takes the asyncio Task that was active at `attach()` time and checks `asyncio.current_task()` against it *before* attempting the detach — a deterministic skip on mismatch, not a try/except around the failure. Catching-and-swallowing the `ValueError` (an earlier version of this fix) hid the problem rather than addressing it: detaching from the wrong Task never resets anything real, because the attaching Task's `Context` object isn't reachable from the finalizer Task in the first place — so attempting it was always a no-op dressed up as an operation. Skipping it explicitly documents that constraint instead of relying on exception-driven control flow. The attaching Task's own context reverts naturally once that Task completes, since contextvars are scoped per-Task and never shared across Tasks.

Scope: `sdk/python/src/openlit/__helpers.py` (new `safe_detach` helper) and `sdk/python/src/openlit/instrumentation/anthropic/{anthropic.py,async_anthropic.py}` (all 4 attach/detach call sites — sync `wrapper()`, async `wrapper()`, `TracedMessageStreamManager`, `TracedAsyncMessageStreamManager`). The sync path keeps a plain try/except fallback since sync generator finalization doesn't cross asyncio Task boundaries the same way.

Regression tests added in `sdk/python/tests/test_context_detach.py`:
- bare cross-Context token mismatch is swallowed (no `attaching_task` hint)
- same-Task exit still performs a real detach (no regression to the happy path)
- foreign-Task exit skips the detach deterministically, without ever attempting it
- end-to-end: `TracedAsyncMessageStreamManager.__aenter__`/`__aexit__` driven from two different asyncio Tasks, mirroring the production traceback, does not raise

### Related issues / prior art

- openlit/openlit#1369 — this bug, reported against the Anthropic instrumentation.
- openlit/openlit#861 — same bug class (`Failed to detach context`), reported against the LangChain/LangGraph instrumentation. Closed, but no merged PR could be found that actually fixes it (`git log --all --grep=861` and history on `instrumentation/langgraph/*` show nothing). Not fixed in this PR — left as documented follow-up since it's the same `context_api.attach`/`detach` idiom, just a different provider module.
- [agno-agi/agno#5208](https://github.com/agno-agi/agno/issues/5208) — same underlying failure signature (context/token detached in a different asyncio Task than it was attached in), observed independently in `openinference-agno-instrumentor` (not an OpenLIT issue, corroborating evidence this is a recurring OTel-instrumentation footgun, not something specific to our wrapper).
- [traceloop/openllmetry `callback_handler.py#L208`](https://github.com/traceloop/openllmetry/blob/main/packages/opentelemetry-instrumentation-langchain/opentelemetry/instrumentation/langchain/callback_handler.py#L208) — another independent instance of this same failure mode in a LangChain instrumentation, using a "catch and ignore" workaround (referenced in the #861 discussion). This PR goes a step further by detecting the mismatch up front instead of catching the resulting error.

I don't have verified issue links for Arize Phoenix or a LangChain-repo-native (not openllmetry) report of this exact failure — didn't want to guess/fabricate URLs. If you have specific issue numbers for those, happy to add them.

### Checklist

* [ ] Changes are documented

### Acknowledgment

## Summary by Sourcery

Guard OpenTelemetry context detaches in Anthropic streaming paths to avoid cross-task context errors and add regression coverage for the new behavior.

Bug Fixes:
- Prevent ValueError and 'Failed to detach context' errors when detaching OTel context tokens from Anthropic async streams finalized in a different asyncio Task.
- Ensure synchronous Anthropic streaming paths detach context tokens safely without propagating detach errors.

Enhancements:
- Introduce a shared safe_detach helper to centralize defensive OTel context detach logic across instrumentation code.

Tests:
- Add regression tests covering safe_detach behavior across same-task, cross-context, and cross-task scenarios, including an end-to-end async Anthropic stream manager case.


ron-42:
@sourcery-ai review

ron-42:
Hi @AmanAgarwal041, please review this pr asap!

amanagarwal042 (approved):

Issue #1298, opened 2026-06-28
Accessibility: improve color contrast, typography, and interactive states


AmanAgarwal041:
## Problem

Several UI areas need an accessibility pass to improve readability, contrast, and interaction affordances across light and dark mode. This is especially visible in navigation, secondary sidebars, feature headers, tables, notices, and compact controls.

## Scope

Audit and fix accessibility issues across the OpenLIT app UI, including:

- Color contrast in light and dark mode
  - Text on muted backgrounds
  - Icons on tinted surfaces
  - Disabled and secondary states
  - Borders/dividers in dark mode
  - Alert/notice backgrounds and foreground colors
- Typography and sizing
  - Minimum readable font sizes for labels, table cells, sidebar items, helper text, badges, and form hints
  - Consistent line-height for dense UI sections
  - Avoid overly small uppercase labels where they reduce readability
- Focus and keyboard states
  - Visible focus rings for buttons, links, tabs, toggles, sidebar items, command/search items, and table actions
  - Keyboard navigation in primary/secondary sidebar menus
  - Escape/click-away behavior for overlays and panels
- Interactive states
  - Hover/active/selected states with sufficient contrast
  - Clear selected states in navigation and tabs
  - Consistent disabled states that remain legible
- Layout/readability
  - Avoid cramped spacing in notices and dense sidebars
  - Ensure compact UI remains readable at common viewport sizes
  - Confirm collapsed navigation tooltips and icons are accessible

## Acceptance criteria

- UI colors meet WCAG AA contrast targets where applicable.
- Text sizes and line heights are readable in both light and dark mode.
- Keyboard users can identify focus and operate navigation panels reliably.
- Sidebar, tables, feature headers, settings pages, and notices have consistent accessible states.
- No regressions in existing light/dark mode styling.

## Suggested validation

- Manual audit in light and dark mode.
- Keyboard-only pass over sidebar, secondary menu, settings, tables, tabs, and dialogs.
- Run an automated accessibility checker such as axe/Lighthouse on representative pages.
- Verify at least these pages:
  - /home
  - /telemetry
  - /prompt-hub
  - /vault
  - /openground
  - /evaluations/settings
  - /settings/profile
  - /settings/api-keys
  - /organisation

AmadNaseem:
[...]

PR #1372, opened 2026-07-16
fix(client): improve keyboard focus visibility

Noone9029:
## Summary

Adds visible keyboard focus states to shared buttons and sidebar links.

## Changes

- adds a consistent focus-visible ring to the shared Button primitive
- adds matching focus styles to primary and secondary sidebar links
- adds focus styles to expanded and compact My Apps links
- preserves existing hover, active, light-mode, and dark-mode behavior

## Motivation

Part of #1298.

Several interactive elements either removed the native browser outline without
providing a replacement or only exposed hover/active states. This made keyboard
focus difficult to identify.

This PR is intentionally limited to keyboard-focus visibility and does not
overlap with the color-contrast work discussed on the issue.

## Validation

- [x] TypeScript typecheck
- [x] Lint
- [ ] Jest suite
- [x] Production build
- [x] Keyboard-only navigation in expanded sidebar
- [x] Keyboard-only navigation in collapsed sidebar
- [x] Secondary panel navigation and Escape behavior
- [x] Light mode
- [x] Dark mode

## Visual evidence

Before:

After:


Noone9029:
I kept the focus bundle inline intentionally: #1298's agreed scope specifies this exact bundle in only the shared Button primitive plus the sidebar and My Apps link paths. Extracting a new shared utility or primitive would broaden the PR beyond that narrowly reviewed accessibility slice. The three call sites now use the same bundle verbatim, while preserving the existing variants, APIs, hover, active, sizing, and tooltip behavior.

amanagarwal042 (approved):
Thanks @Noone9029 , Very much needed and very well done PR. Looking forward to future PRs and issues from you!!

PR #1354, opened 2026-07-12
feat: add a dimension registry for trace analysis

morgan-coded:
> [!IMPORTANT]  
> 1. We strictly follow a issue-first approach, please first open an [issue](https://github.com/openlit/openlit/issues) relating to this Pull Request.
> 2. PR name follows conventional commit format: `feat: ...` or `fix: ....`

**Issue number**: #1327

### Change description:

Trace analysis currently has no single config layer for its six analysis dimensions, which makes focused additions or removals touch several separate paths. This adds the registry-first layer while keeping the current six dimensions, prompts, labels, order, and stored-run JSON unchanged. New dimensions or removals stay out of this first pass and can follow as focused PRs.

### Checklist

If your change doesn't seem to apply, please leave them unchecked.
* [x] PR name follows conventional commit format: `feat: ...` or `fix: ....`
* [x] I have reviewed the [contributing guidelines](https://github.com/openlit/openlit/blob/main/CONTRIBUTING.md)
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/openlit/openlit/pulls) for the same update/change?
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented

### Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/openlit/openlit/blob/main/LICENSE).

Issue #1327, opened 2026-07-03
Feature: Expand AI analysis into configurable dimensions for trace hierarchies and individual spans


AmanAgarwal041:
### Component

OpenLIT

### What feature would you like to see?

Expand the AI analysis experience so it can evaluate traces and spans across more configurable dimensions, not just the current improvement-oriented set.

The current analysis is useful, but it is still too narrow for broader operational review. We should support a richer and more modular set of dimensions such as:

- [ ] Guardrails and policy adherence
- [ ] Security issues and risky behavior
- [ ] PII or sensitive-data exposure
- [ ] Prompt injection and jailbreak signals
- [ ] Hallucination or unsupported-claim risk
- [ ] Cache efficiency and cache-improvement opportunities
- [ ] Tool misuse or suspicious tool-call patterns
- [ ] Latency bottlenecks and slow spans
- [ ] Cost and token waste
- [ ] Retry amplification and error loops
- [ ] Context bloat or unnecessary payload growth
- [ ] Missing validation or weak control-flow checks
- [ ] Hierarchy-level dependency issues across parent/child spans
- [ ] Anomaly detection and unusual behavior compared with neighboring spans
- [ ] Severity, confidence, and prioritization scoring

This should work at two levels:

- trace hierarchy analysis
- individual span analysis

The key requirement is modularity. People should be able to pick up a single dimension at a time, or enable a focused subset, rather than being forced to run the full analysis suite every time.

Create a sub issue and pick up individual dimensions at once.

### Why do you need this?

A broader analysis model would make OpenLIT more useful for debugging, security review, observability, and optimization. Different users care about different signals: some want guardrails and compliance, some want performance and cost, and some want root-cause analysis across the trace tree. A configurable analysis system would let users focus on the dimensions that matter for their workflow while keeping the output concise and actionable.

### Additional context

A good implementation would likely include:

- A dimension registry or config layer so analysis categories can be added without rewriting the whole flow
- Support for running one dimension, many dimensions, or all dimensions
- Trace-level and span-level support using the same underlying dimension model
- Clear output grouping by dimension with summaries and findings
- Optional severity and confidence metadata for prioritization
- Backward compatibility with the current AI analysis output shape, or a migration path for existing stored runs

## Files likely involved

- `src/client/src/types/trace-analysis.ts`
- `src/client/src/lib/platform/chat/improvement.ts`
- `src/client/src/components/(playground)/request/components/trace-improvement-view.tsx`
[...]

PR #1376, opened 2026-07-16
feat(client): add prompt-injection/jailbreak trace-analysis dimension

### LINKED ISSUE #1367: Add prompt injection and jailbreak signals to trace analysis
This is the sub-issue #1327 requests for prompt injection and jailbreak signals, building on the merged trace-analysis registry from #1354.

I'm claiming only this dimension for the first slice. If you'd rather start with a different first dimension, I can switch.

### Scope

I'll add the dimension definition to `TRACE_ANALYSIS_DIMENSION_DEFINITIONS` and keep evidence selection scoped to `systemPrompt`, `prompt`, `response`, `toolName`, `toolCallId`, `toolArgs`, and `toolResult`. The localized label, guidance, and empty-state copy will live in `en.ts` from the start.

Older stored analysis should remain readable when the new key is absent. My default is to normalize missing keys on read; if you'd prefer UI-boundary handling, I'll use that contract instead.

### Tests

I'll cover registry order and freezing, exact field selection, compatibility with old runs, parser rejection of findings attributed to another dimension, mocked trace and span flows, and both UI empty states. I'll also keep a regression around the existing dimensions and stored-result behavior.

### Non-goals

Other dimensions stay separate, one at a time as #1327 requests. This slice will not include database schema changes, a UI redesign, or runtime prompt-injection prevention.

morgan-coded:
Adds the prompt-injection/jailbreak signals dimension from #1367 — the first security slice of #1327 — through the trace-analysis registry from #1354.

Scope, per my claim on #1367 (this dimension only for the first slice):
- One `defineDimension` entry in `TRACE_ANALYSIS_DIMENSION_DEFINITIONS`: guidance targets instruction-override, role-confusion, and jailbreak-framing evidence in prompt/completion content and tool-call arguments; spanFields/metricFields project through the existing selectors — no bespoke UI or parallel analysis path.
- Key union + registry map entries; all user-facing copy in `en.ts` (`TRACE_AI_*`).
- Missing-field handling: spans without content fields normalize to empty evidence rather than erroring — stated here as a reviewable choice since #1367's normalization question is still open; happy to flip it.

Tests extend `trace-analysis-registry.test.ts`: definition-shape completeness, selector projection on injection-bearing vs clean span fixtures, the missing-field case, and en.ts key presence.

Worth stating plainly: this is an observability signal, not a guardrail — the copy avoids any detection-guarantee language.

Non-goals: tool-misuse and retry-amplification (separate #1327 checkboxes — I'll file the tool-misuse sub-issue next, per the umbrella's workflow), severity/confidence scoring, and any change to the six existing dimensions.

Refs #1327, closes #1367.

## Summary by Sourcery

Add a new prompt-injection/jailbreak trace-analysis dimension and integrate it into the existing registry and messaging system.

New Features:
- Introduce a prompt_injection trace-analysis dimension that surfaces prompt and tool-call content relevant to instruction override, role confusion, and jailbreak framing.

Enhancements:
- Extend trace span selection to optionally normalize missing content fields to empty strings for dimensions that represent observability evidence.
- Wire the prompt_injection dimension into the trace-analysis dimension registry with appropriate labels, guidance, and empty-state copy from the English message table.

Tests:
- Expand trace-analysis registry tests to cover the prompt_injection dimension definition, selector behavior on injection-bearing versus clean spans, missing-field normalization, and presence of all related en.ts message keys.


amanagarwal042:
Thanks @morgan-coded , Ill take a loot at this

amanagarwal042 (approved):

PR #1379, opened 2026-07-18
fix(cost): price cached tokens at cache rates in litellm and claude_agent_sdk

### LINKED ISSUE #1378: Bug: litellm and claude_agent_sdk bill Anthropic cache-read tokens at full prompt price
### Component

OpenLIT Python SDK

### What happened

The `litellm` and `claude_agent_sdk` instrumentations record request cost by passing an input-token count that already includes Anthropic cache tokens to `get_chat_model_cost`, but without the cache arguments. Cache-read tokens are therefore billed at the full prompt price instead of `cacheReadPrice` (one tenth of the prompt price for Claude models), inflating the reported `gen_ai.usage.cost`.

Both adapters build a cache-inclusive input total:

- `claude_agent_sdk/utils.py`: `extract_usage` sets `input_tokens = uncached_input + cache_read + cache_creation`, then `set_chat_span_attributes` passes only that inclusive total to `_calculate_cost`.
- `litellm/utils.py`: `common_chat_logic` uses litellm's normalized `prompt_tokens`, which litellm reports inclusive of cache-read and cache-creation tokens.

Separately, the litellm adapter reads cache-creation tokens from `usage["completion_tokens_details"]["cached_tokens"]`, a field litellm does not populate, so `gen_ai.usage.cache_creation_input_tokens` is always 0.

### Steps to reproduce

Measured offline against `main` (104e6874) in `python:3.12-slim` with no network, driving each adapter's own code over a usage object of 200 uncached input, 5000 cache-read, 1000 cache-creation, 300 output on a Claude model priced (per 1K) at promptPrice 0.003, cacheReadPrice 0.0003, cacheCreationPrice 0.00375, completionPrice 0.015:

```python
# claude_agent_sdk
attrs = extract_usage({"input_tokens": 200, "cache_read_input_tokens": 5000,
                       "cache_creation_input_tokens": 1000, "output_tokens": 300})
# attrs input_tokens == 6200  (inclusive)
_calculate_cost(model, pricing, 6200, 300)          # -> $0.023100

# litellm: AnthropicConfig().calculate_usage(...) normalizes prompt_tokens to 6200
get_chat_model_cost(model, pricing, 6200, 300)      # -> $0.023100
```

Ground truth for that request (uncached at prompt price, cache-read at cacheReadPrice, cache-creation at cacheCreationPrice) is `$0.010350`, so the reported cost is 2.23x too high. Because promptPrice is 10x cacheReadPrice for the shipped Claude rates, the overcharge grows with the cached fraction of the prompt.

### Expected behavior

Cache-read tokens are billed at `cacheReadPrice` and cache-creation tokens at `cacheCreationPrice`, so `gen_ai.usage.cost` matches the provider's actual bill, and `gen_ai.usage.cache_creation_input_tokens` reflects the real cache-write count.

### Environment

- OpenLIT Python SDK: `main` @ 104e6874
- litellm: 1.92.0 (used only to confirm normalization; not needed to reproduce the arithmetic)
- Python: 3.12

ebarkhordar:
**Issue number**: #1378

Fixes #1378

### Change description

The `litellm` and `claude_agent_sdk` instrumentations pass a cache-inclusive input-token count to `get_chat_model_cost` without the cache arguments, so cache-read tokens are billed at the full prompt price instead of `cacheReadPrice`. For Claude models (the only models in `assets/pricing.json` that carry `cacheReadPrice`, all at 10x the read rate) this inflates `gen_ai.usage.cost` in proportion to the cached fraction of the prompt.

Root cause:

- `claude_agent_sdk/utils.py`: `extract_usage` builds `input_tokens = uncached + cache_read + cache_creation`, and `set_chat_span_attributes` forwarded only that inclusive total to `_calculate_cost`.
- `litellm/utils.py`: `common_chat_logic` forwarded litellm's normalized `prompt_tokens`, which is inclusive of cache-read and cache-creation tokens, without the cache arguments.

Invariant: when an adapter's `prompt_tokens` already includes cache tokens, the cost call must forward the cache counts with `prompt_tokens_include_cache=True`, so the shared helper re-prices those tokens at `cacheReadPrice`/`cacheCreationPrice` and subtracts them from the prompt base instead of billing them at prompt price. `get_chat_model_cost` already implements this; only the two call sites were missing the arguments.

Second defect in litellm: cache-creation tokens were read from `completion_tokens_details.cached_tokens`, a field litellm never populates, so the count was always 0. litellm reports it under `prompt_tokens_details.cache_creation_tokens`. This is now read from the correct field via a small `_extract_cache_tokens` helper that also de-duplicates the parsing shared by the streaming and non-streaming paths.

Behavioral change beyond cost: fixing the litellm cache-creation source means the `gen_ai.usage.cache_creation_input_tokens` span attribute now reports the real cache-write count instead of 0. Cost for non-Claude models is unchanged, since `get_chat_model_cost` only re-prices cache tokens when the model defines `cacheReadPrice`/`cacheCreationPrice` (no non-Claude model in the bundled pricing does), so `prompt_tokens_include_cache=True` is inert for them.

Scope: this covers the two adapters I could verify build an inclusive total and route Claude models. `strands/processor.py:417` also omits the cache arguments, but its `gen_ai.usage.input_tokens` comes straight from the external Strands SDK and openlit does not sum cache tokens into it, so whether it is cache-inclusive depends on that SDK. I could not verify that offline and left it out rather than guess.

### How this was verified

In a clean `python:3.12-slim` container, no network, against this branch, with each module's `__file__` and sha256 asserted equal to the checkout:

- Drove `claude_agent_sdk` `extract_usage` then `_calculate_cost`: inclusive total 6200, fixed cost `$0.010350` == ground truth (`main` billed `$0.023100`, 2.23x).
- Confirmed litellm's own `AnthropicConfig().calculate_usage` (litellm 1.92.0) returns `prompt_tokens = 6200` for `{input 200, cache_read 5000, cache_creation 1000}` and exposes `prompt_tokens_details.cache_creation_tokens = 1000`, while `completion_tokens_details` carries no cache count.
- `tests/test_cache_pricing.py`: 16 passed, 1 skipped on this branch. The two new regression tests (`TestClaudeAgentSDKAdapter::test_cost_reprices_cache_tokens` and `TestLiteLLMAdapter::test_extract_cache_tokens_reads_prompt_tokens_details`) fail on `main` and pass here.
- Ran the existing `test_claude_agent_sdk.py`, `test_litellm_reproduction.py`, `test_reasoning_content.py`: identical results with and without this change (160 passed; 8 pre-existing, environment-dependent failures unrelated to this diff).
- `pylint` 4.0.6 rates the three changed files 10.00/10.

The new tests are dict-driven and import neither `litellm` nor `claude_agent_sdk` (neither is a dependency). `python-tests.yml` is disabled (`.hold`) in this repo, so the SDK tests do not run in CI; they were run locally as above.

### Checklist

* [x] PR name follows conventional commit format: `fix: ...`
* [x] Changes are documented

### Acknowledgment


amanagarwal042 (approved):
This is a great fix @ebarkhordar , Really appreciate the changes. Looking forward to more PRs and issues from you!

ebarkhordar:
Thanks @amanagarwal042. The token counts were already correct on the span; only the cost call read the cache-inclusive input total at the full prompt rate, so the fix was just forwarding the cache args and the include-cache flag at those two sites. Glad it helped, and I will keep sending fixes as I find them.

PR #1381, opened 2026-07-18
fix(python-sdk): compute AssemblyAI cost from audio duration

JoshKappler:
## Problem

`common_audio_logic` in the AssemblyAI instrumentation passes the transcript `audio_url` as the `prompt` argument to `get_audio_model_cost`:

```python
prompt = scope._response.audio_url
cost = get_audio_model_cost(request_model, pricing_info, prompt, scope._response.audio_duration)
```

`get_audio_model_cost` uses the prompt when it is truthy and only falls back to duration otherwise:

```python
if prompt:
    cost = (len(prompt) / 1000) * pricing_info["audio"][model]
else:
    cost = duration * pricing_info["audio"][model]
```

`audio_url` is always set on a successful transcription, so the duration branch is never reached and `gen_ai.usage.cost` comes out of the character length of the URL. The `best` and `nano` entries in `assets/pricing.json` are per-second rates (0.00010277777 * 3600 = $0.37/hr), so a 2 hour transcription submitted as a 38 character URL reported $0.0000038 instead of $0.74. Re-uploading the same audio behind a longer presigned URL changed the reported cost.

## Fix

Pass `None` and the duration, which is what the groq (`groq/utils.py:861`) and openai (`openai/utils.py:2399`) audio instrumentors already do. `audio_url` is still used for `build_input_messages`, so captured input content is unchanged.

## Verification

Added `sdk/python/tests/test_assemblyai_cost.py` covering cost from duration, cost independence from URL length, the nano rate, unknown models, and that the URL is still recorded as input content. The duration and URL-length tests fail on the current code and pass with this change.

```
python -m pytest tests/test_assemblyai_cost.py -q   -> 5 passed
python -m ruff format --check / ruff check          -> clean
python -m pylint <both files>                       -> 10.00/10
```

## Summary by Sourcery

Compute AssemblyAI transcription cost from audio duration instead of URL-derived prompt length.

Enhancements:
- Align AssemblyAI audio cost calculation with per-second billing by basing it on transcript duration rather than audio URL length.

Tests:
- Add tests verifying AssemblyAI cost is derived from duration, independent of URL length, respects per-model rates, handles unknown models, and still records the audio URL as input content.


amanagarwal042:
Thanks @JoshKappler for this fix. Really appreciate it. Looking forward to more PRs and issues from you!!

Issue #1353, opened 2026-07-12
Bug: Helm chart repository missing published versions 1.22.0/1.23.0; security findings on 1.20.0 image and ClickHouse dependency


sqrt676:
### Component

OpenLIT

### What happened?

Labels: bug, security, helm-chart
**Description:**
Summary While evaluating OpenLIT for deployment, we found inconsistencies in the published Helm repository and security concerns in the current deployable version.
**1. Helm chart version mismatch**
    •    Only 1.20.0 is currently available and pullable from the public Helm repository.
    •    Versions 1.22.0 and 1.23.0 are referenced (e.g., in docs/changelog) but are not present in the published Helm repo index, so helm pull/helm install fails for these versions.
    •    This makes 1.20.0 the only practical option for deployment/evaluation today.

**2. Security findings on OpenLIT image 1.20.0 (via Wiz scan)**
    •    Fails repo-CICD vulnerability policy
    •    Multiple Critical/High severity vulnerabilities
    •    Ships with Node.js 18 and Next.js 14 — both EOL
    •    Where testable, 1.22.0/1.23.0 images show similar vulnerability/EOL profiles, so no clear security improvement over 1.20.0
**3. ClickHouse dependency (pinned to 24.4.1)**
    •    The 1.20.0 Helm chart pins ClickHouse 24.4.1
    •    Base image uses Ubuntu 20.04, which is EOL — flagged as a Critical finding
    •    Remaining findings are OS-package level, not application-level
**4. ClickHouse 26.3 LTS compatibility**
    •    We separately validated ClickHouse 26.3 LTS — onboards cleanly, no comparable EOL issue
    •    However, current Helm chart/deployment artifacts only reference 24.4.1; there’s no documented support/compatibility for 26.3 with OpenLIT 1.20.0
**Ask**
    1    Please publish/fix the Helm repo index so 1.22.0/1.23.0 are actually pullable
    2    Confirm plan to address Node.js 18 / Next.js 14 EOL dependencies
    3    Confirm whether OpenLIT 1.20.0 (or later) can be validated/supported against ClickHouse 26.3 LTS

### Steps to reproduce

**Steps to reproduce**
    1    Add the OpenLIT Helm repo
    2    Attempt helm pull openlit --version 1.22.0 (or 1.23.0)
    3    Observe failure — chart version not found in index

### Expected behavior

pass the wiz scan ideally

### Pre-submission checklist

- [x] I searched existing issues and didn't find a duplicate

[...]

PR #1391, opened 2026-07-22
fix(client): upgrade Node 18→22 and Next 14→16 to clear EOL/CVEs (#1353)

bhavik168:
**Issue number**: #1353 (Part 2 of 4 — Node.js / Next.js EOL)

### Change description:

Addresses **Part 2** of #1353: Node.js 18 and Next.js 14 are both EOL and were the main Wiz/CVE findings. Parts 3 & 4 (ClickHouse) will follow as separate PRs. This is a large change — the React 19 upgrade touches ~57 files — so here's a summary.

**1. Runtime & framework**
- Node.js **18 → 22** (LTS) and base image **Debian 11 → 12 (bookworm)** — both were EOL (the top "Critical" hits).
- Next.js **14.2.4 → 16.2.11** and React **18 → 19**. Staying on 14 left a CRITICAL 9.1 middleware auth-bypass and several HIGH CVEs permanently unpatched (14 is EOL; fixes only exist in ≥15.5.16). Next's own advisories are now cleared.

**2. Dependency changes forced by React 19**
- `react-beautiful-dnd` (archived) → `@hello-pangea/dnd` (drop-in, React-19-compatible fork).
- `cmdk` → 1.1.1 and `@textea/json-viewer` → v4 (React-19-compatible MUI 6) for type/runtime compatibility.
- Added a documented `.npmrc` (`legacy-peer-deps`) because some UI deps still declare a `react ^17||^18` peer despite working with 19. Since that disables peer auto-install, previously-transitive peers were promoted to explicit deps: `@mui/material`, `@emotion/react`, `@emotion/styled`, `zod`, `monaco-editor`, `@testing-library/dom`.

**3. Code migrations (bulk of the diff)**
- **Sync → async request APIs** (Next 15+ made these Promises): migrated ~33 route handlers and pages to await `params`, and awaited `cookies()` in the root layout. Used the official `next-async-request-api` codemod, then hand-fixed the wrapper-pattern handlers it couldn't follow.
- **React 19 type fixes**: `children: JSX.Element` → `ReactNode`, global `JSX.*` → `ReactNode`, `useRef<T>()` now requires an initial arg, explicit typing for Monaco callbacks.
- **Config**: removed the removed `experimental.instrumentationHook`; migrated ESLint 8 → 9 with a flat `eslint.config.mjs` (`next lint` is removed in Next 16, so the script is now `eslint .`).

**Verification** — full `CLAUDE.md` contributor gate passes locally and in a clean Docker build (Node 22 image):
- `next build` (Next 16 / Turbopack) — exit 0, 159 pages
- `npm test -- --runInBand` — 216 suites / 2661 tests pass
- `npm run lint` — exit 0
- `npx prisma validate` — valid
- Docker image builds end-to-end (clean-container install + prisma generate + next build)

**Note:** one residual `sharp`/libvips HIGH remains — a pre-existing dependency unrelated to Node/Next, a trivial separate bump. The `middleware` → `proxy` rename (a Next 16 *deprecation*, not an error) is intentionally left as a follow-up since it touches auth-routing.

### Checklist

* [ ] Changes are documented

### Acknowledgment


src/client/src/app/(playground)/organisation/project/[projectId]/page.tsx
```diff
@@ -1,9 +1,10 @@
 import OrganisationProjectPage from "@/components/(playground)/organisation/project-page";
 
-export default function ProjectRoutePage({
-	params,
-}: {
-	params: { projectId: string };
-}) {
-	return <OrganisationProjectPage projectId={params.projectId} />;
```
Comment (AmanAgarwal041) [inline src/client/src/app/(playground)/organisation/project/[projectId]/page.tsx:8]: Follow the same pattern, either use or await props.params. Don't use both , one in some files and other method in other when refactoring. Also specify why `use` is important ?

src/client/package.json
```diff
@@ -22,9 +22,13 @@
     "@anthropic-ai/sdk": "^0.21.1",
     "@clickhouse/client": "^1.1.0",
     "@dhmk/zustand-lens": "^4.0.0",
+    "@emotion/react": "^11.11.4",
+    "@emotion/styled": "^11.11.5",
+    "@hello-pangea/dnd": "^18.0.1",
     "@kubernetes/client-node": "^1.4.0",
     "@mistralai/mistralai": "^0.4.0",
     "@monaco-editor/react": "^4.7.0",
+    "@mui/material": "^6.5.0",
```
AmanAgarwal041, line 31: Why is this required ? when we already have tailwind. Also why add new packages

AmanAgarwal041:
@bhavik168 Thanks for the contribution, please address the comments and update your branch from the main.

src/client/src/app/(playground)/organisation/project/[projectId]/page.tsx
Comment (bhavik168) [reply src/client/src/app/(playground)/organisation/project/[projectId]/page.tsx:8]: Actually, as Next.js now gives us `params` as a promise, we have to unwrap it before reading the id. Server pages can be `async`, so they just `await props.params` that's 49 of the 50 files. But `openground/[id]/page.tsx` has `"use client"` (it uses browser stuff like analytics and `useEffect`), and browser components can't be `async`, so `await` isn't used there instead `use()` which is React's way to unwrap the promise is implemented

If you'd rather have every file use `await`, I can move that page's params handling into a tiny server wrapper and pass the id down. Should I go for it?

src/client/package.json
bhavik168: We're not actually using this UI library, none of our own code imports MUI, Tailwind still does all the styling. `@mui/material` (and the two `@emotion` packages) are behind-the-scenes dependencies of `@textea/json-viewer`, the JSON viewer we already use.

They used to install automatically. But the React 19 upgrade needed `legacy-peer-deps` turned on in `.npmrc`, and that setting stops npm from auto-installing these background packages, so they have to be listed explicitly now, otherwise the install and Docker build fail. Same reason `zod`, `monaco-editor`, and `@testing-library/dom` are listed too.

### Any suggestion from your side what should we do here?

Issue #1412, opened 2026-07-28
Bug: openlit deployment in air-gaped environments


JoseAlbanoGaspar:
### Component

OpenLIT

### What happened?

Problem: Dockerfile

First stage (Alpine container), the npm install and prisma generate commands will install:

- schema engine for alpine only (cause the container is alpine - this is a known behaviour for prisma [here](https://blog.tericcabrel.com/build-docker-image-nodejs-prisma/))
- query engine for multiple environments (alpine, debian, etc)

The final image is debian.

Result: without internet access (air gaped environments), the container hangs cause prisma detects there is only a schema engine for alpine, and not for debian. Works fine with internet access cause prisma can download the right version afterwards.

### Steps to reproduce

1. Install openlit via helm chart and remove internet access to the cluster

### Expected behavior

All prisma dependencies should be created during image build phase and not runtime.

### Pre-submission checklist

- [x] I searched existing issues and didn't find a duplicate

### Are you willing to submit PR?

None

PR #1414, opened 2026-07-28
fix(client): bake glibc Prisma engines at build time to avoid launch-time network calls

jpv-costa:
**Issue number**: #1412

### Change description:

Fixes the platform container hanging/failing at launch in air-gapped clusters because Prisma downloads an engine from `binaries.prisma.sh` at startup.

**Root cause**: `src/Dockerfile` installs dependencies and generates the Prisma client in an **Alpine 3.21 (musl)** builder stage, then copies `node_modules/@prisma`, `node_modules/.bin`, and `node_modules/prisma` into the **`node:18-bullseye-slim` (Debian, glibc, OpenSSL 1.1.x)** runtime stage. `schema.prisma` `binaryTargets` only governs the **Query Engine** embedded in the generated client, so it correctly ships a `debian-openssl-1.1.x` query engine. It does **not** govern the **CLI/Schema Engine** used by `prisma migrate deploy` and `prisma generate` — that binary is resolved at install time for the *builder* platform only, so the image contains just `schema-engine-linux-musl-openssl-3.0.x`. On the glibc runtime that binary can't execute, so at launch the Prisma CLI in `entrypoint.sh` reaches out to `binaries.prisma.sh` to download a matching engine. With no egress the request fails (`getaddrinfo EAI_AGAIN binaries.prisma.sh`) and the container never starts. (Matches the analysis in #1412.)

**Fix**: set `PRISMA_CLI_BINARY_TARGETS=linux-musl-openssl-3.0.x,debian-openssl-1.1.x` in the builder stage before `npm install`, so the runtime (Debian/glibc) CLI engines are fetched at **build time** and baked into the image alongside the musl builder engines. No entrypoint or `schema.prisma` change is needed — `prisma migrate deploy` and `prisma generate` then run fully offline against the baked engines.

Notes:
- The runtime is Debian **bullseye** = OpenSSL **1.1.x**, so the correct target is `debian-openssl-1.1.x` (not `-3.0.x`).
- `PRISMA_CLI_BINARY_TARGETS` requires explicit platform names; the builder's own (musl) target must stay in the list or the build-time `prisma generate` fails with `Unknown binaryTarget`.

**Scope**: `src/Dockerfile` only (one `ENV` line + explanatory comment). Runtime behavior is otherwise unchanged: migrations still applied, seed still runs, `PrismaClient` queries, OpAMP/OTel (localhost), NextAuth, and cron are untouched.

### Verification

Reproduced with a minimal build using Prisma 5.15.0 and this repo's real `schema.prisma` + `prisma/migrations`, mirroring the Alpine-builder → `node:18-bullseye-slim`-runtime copy pattern:

- **Before** (`docker run --network=none`): only `schema-engine-linux-musl-openssl-3.0.x` (interp `/lib/ld-musl-x86_64.so.1`) is baked; both `prisma migrate deploy` and `prisma generate` fail with `Error: request to https://binaries.prisma.sh/.../debian-openssl-1.1.x/... failed, reason: getaddrinfo EAI_AGAIN binaries.prisma.sh`.
- **After** (`docker run --network=none`): `schema-engine-debian-openssl-1.1.x` (interp `/lib64/ld-linux-x86-64.so.2`, glibc) is baked alongside the musl engine; `prisma migrate deploy` applies all migrations ("All migrations have been successfully applied") and `prisma generate` completes — zero outbound requests.
- Built the **real** `src/Dockerfile` builder stage with the change: the full Next.js build succeeds and `node_modules/@prisma/engines/` now contains both `schema-engine-debian-openssl-1.1.x` (glibc) and `schema-engine-linux-musl-openssl-3.0.x`, plus the `debian-openssl-1.1.x` query engine — i.e. the engines copied into the runtime stage are runtime-compatible.

### Checklist

* [ ] Changes are documented

### Acknowledgment

## Summary by Sourcery

Bake platform-compatible Prisma engines into the client image to prevent startup failures in air-gapped deployments.

Bug Fixes:
- Bake Prisma CLI engines for the runtime platforms during the client image build so migrations and client generation work without launch-time network access in air-gapped environments.

Build:
- Configure the client Docker build to include Prisma engines for the supported Alpine and Debian runtime architectures.


src/Dockerfile
```diff
@@ -13,6 +13,19 @@ WORKDIR /app/client
...
+#
+# schema.prisma `binaryTargets` only governs the Query Engine embedded in the
+# generated client. The Schema Engine used by `prisma migrate deploy` (and the
+# CLI engines used by `prisma generate`) are resolved at install time for the
+# *builder* platform only (musl). When those musl engines are copied into the
+# glibc runtime, the Prisma CLI cannot execute them and downloads a matching
+# engine from binaries.prisma.sh at container launch -- which fails in
+# air-gapped / locked-down clusters. Forcing the runtime target here makes
+# `prisma migrate deploy` and `prisma generate` run fully offline at launch.
+ENV PRISMA_CLI_BINARY_TARGETS=linux-musl-openssl-3.0.x,debian-openssl-1.1.x
```
AmanAgarwal041, line 27: @jpv-costa  Can you update the line to : 
```
# Recommended (x64 + ARM64):

ENV PRISMA_CLI_BINARY_TARGETS=linux-musl-openssl-3.0.x,debian-openssl-1.1.x,linux-musl-arm64-openssl-3.0.x,linux-arm64-openssl-1.1.x
```

amanagarwal042:
Hey @jpv-costa Thanks for the PR!
Can you also reach out to me over email (contact@openlit.io), I have a proposal for you and your team

Issue #1400, opened 2026-07-24
Bug: Gemini (google_ai_studio) instrumentor drops parallel function_call parts


AmirF194:
### Component

OpenLIT Python SDK

### What happened?

The Gemini (`google_ai_studio`) instrumentor only ever reads the first `function_call` part of a candidate's content, so a turn with more than one parallel function call silently loses the rest.

- **Non-streaming** (`process_chat_response` in `sdk/python/src/openlit/instrumentation/google_ai_studio/utils.py`): `fc = parts[0].get("function_call") if parts else None` inspects `parts[0]` only. A second or third `function_call` part in the same candidate is never read.
- **Streaming** (`process_chunk`): the same `parts[0]` read is assigned straight into `scope._tools` on every chunk, overwriting the previous value. Gemini can split parallel function calls across separate streamed chunks, so even the first call is lost once a later chunk carries the second.

### Steps to reproduce

1. Instrument a Gemini (`google-genai`) client with OpenLIT and send a request whose response contains two `function_call` parts in the same candidate (Gemini's documented parallel function calling), whether in one response or split across two streamed chunks.
2. Inspect `gen_ai.output.messages` on the resulting span.

Minimal reproduction of the underlying logic (no live API needed):

```python
candidates = [{"content": {"parts": [
    {"function_call": {"name": "get_weather", "args": {"city": "nyc"}}},
    {"function_call": {"name": "get_weather", "args": {"city": "sf"}}},
]}}]

parts = candidates[0]["content"]["parts"]
fc = parts[0].get("function_call") if parts else None
print(fc)  # {'name': 'get_weather', 'args': {'city': 'nyc'}} -- the second call is never read
```

### Expected behavior

Every `function_call` part in a candidate should be captured, and streamed calls that arrive across multiple chunks should accumulate rather than overwrite. `gen_ai.output.messages` should carry one `tool_call` part per function call.

### Environment

- OpenLIT SDK version: HEAD (dbb8373a7cb539fb6725b1c9687c7dd9621be975)
- Python version: 3.11

### Additional context

Same root cause and shape as the sibling Anthropic bug (#1398): an instrumentor built for a single tool call per turn, not updated for the provider's parallel tool-calling.

Kayvan-Zahiri:
PR #1417 is up for this.

[...]

PR #1416, opened 2026-07-29
fix(anthropic): keep every parallel tool_use block on the span

### LINKED ISSUE #1398: Bug: Anthropic instrumentor corrupts span on parallel tool_use blocks
### Component

OpenLIT Python SDK

### What happened?

The Anthropic instrumentor corrupts the span when a single turn contains more than one parallel `tool_use` block (Claude's parallel tool calling).

- **Streaming** (`process_chunk` / `process_streaming_chat_response` in `sdk/python/src/openlit/instrumentation/anthropic/utils.py`): the tool id and name are tracked as scalars, overwritten on every `content_block_start`, and every block's `partial_json` delta is appended into one shared string. With two tool calls, the recorded id/name belong to the *last* call while the recorded arguments are the *concatenation* of both calls' JSON, which is not valid JSON.
- **Non-streaming** (`process_chat_response`): the loop over `content` blocks breaks after the first `tool_use` block is found, so every parallel call after the first is silently dropped from the span entirely.

### Steps to reproduce

1. Instrument an Anthropic client with OpenLIT and send a `messages.create` request whose response contains two `tool_use` content blocks in the same turn (e.g. a prompt that triggers two independent tool calls Claude can run in parallel).
2. Streaming: inspect `gen_ai.tool.id` / `gen_ai.tool.name` / `gen_ai.tool.args` and `gen_ai.output.messages` on the resulting span.
3. Non-streaming: inspect the same attributes on a non-streaming `messages.create` call whose response has two `tool_use` blocks.

Minimal reproduction of the streaming aggregation logic (no live API needed):

```python
from types import SimpleNamespace
from openlit.instrumentation.anthropic import utils as u

scope = SimpleNamespace(_timestamps=[], _start_time=0, _llmresponse="",
                         _tool_id="", _tool_name="", _tool_arguments="")
chunks = [
    {"type": "content_block_start", "index": 0,
     "content_block": {"type": "tool_use", "id": "toolu_01", "name": "get_weather"}},
    {"type": "content_block_delta", "index": 0,
     "delta": {"partial_json": '{"city": "nyc"}'}},
    {"type": "content_block_start", "index": 1,
     "content_block": {"type": "tool_use", "id": "toolu_02", "name": "get_weather"}},
    {"type": "content_block_delta", "index": 1,
     "delta": {"partial_json": '{"city": "sf"}'}},
]
for c in chunks:
    u.process_chunk(scope, c)

print(scope._tool_id)         # toolu_02 -- id of the SECOND call only
print(scope._tool_arguments)  # {"city": "nyc"}{"city": "sf"} -- not valid JSON
```

### Expected behavior

Every parallel `tool_use` block in a turn should be captured intact: its own id, name, and arguments, and all of them present in `gen_ai.output.messages` as separate tool-call parts. The sibling OpenAI instrumentor (`sdk/python/src/openlit/instrumentation/openai/utils.py`) already tracks streaming tool calls in an index-keyed list instead of scalars, which is the correct pattern for this case.

### Environment

- OpenLIT SDK version: HEAD (dbb8373a7cb539fb6725b1c9687c7dd9621be975)
- Python version: 3.11

### Additional context

`anthropic`'s streaming `content_block_start` / `content_block_delta` events carry an `index` field that identifies which content block a delta belongs to; the current code never reads it, which is the root cause of both symptoms above.

Kayvan-Zahiri:
Fixes #1398.

Claude can return several `tool_use` blocks in a single turn, and both Anthropic paths assumed one call per turn.

**Streaming** (`process_chunk`): the tool id and name were scalars overwritten by every `content_block_start`, and each block's `partial_json` was appended into one shared string. With two calls the span recorded the *second* call's id next to *both* calls' JSON concatenated together, which is not valid JSON.

**Non-streaming** (`process_chat_response`): the loop over content blocks `break`s after the first `tool_use` block, so every parallel call after the first was silently dropped.

## Fix

Streaming now accumulates into an index-keyed list using the block `index` Anthropic already sends, which is the same approach the `openai` instrumentor uses for streamed tool calls. Non-streaming collects every `tool_use` block instead of stopping at the first.

`build_output_messages` emits one `tool_call` part per call, so all parallel calls land in `gen_ai.output.messages`. It still accepts a single dict, so any caller passing the previous shape keeps working.

The single-valued OTel tool attributes (`gen_ai.tool.name` / `.call_id` / `.args`) comma-join across calls, matching what the `openai` instrumentor already does. **For a single tool call the emitted attributes are byte-identical to before**, so existing dashboards do not move.

## Verification

The report's streaming repro now gives distinct ids and independently parseable arguments:

```
{'id': 'toolu_01', 'name': 'get_weather', 'input': '{"city": "nyc"}'}
{'id': 'toolu_02', 'name': 'get_weather', 'input': '{"city": "sf"}'}
```

Added `tests/test_anthropic_parallel_tools.py`, offline and needing no API key, covering:

- parallel streaming calls stay separate and each `input` parses as JSON
- one block's arguments split across several deltas still reassemble
- interleaved text blocks do not create phantom tool calls
- non-streaming keeps every `tool_use` block, in order
- the legacy single-dict shape still works
- empty or non-dict entries are skipped

```
6 passed
pylint 10.00/10 (.pylintrc, changed files)
```

The sibling Gemini bug #1400 has the same root cause in `google_ai_studio`; happy to take that one next if this approach looks right.

## Summary by Sourcery

Handle parallel Anthropic tool_use blocks correctly in both streaming and non-streaming instrumentation paths while preserving existing single-call behavior.

New Features:
- Support multiple parallel tool_use blocks per turn in Anthropic instrumentation, emitting one tool_call part per tool invocation in output messages.

Bug Fixes:
- Prevent streamed parallel Anthropic tool_use blocks from overwriting each other’s ids and concatenating their JSON arguments into an invalid single payload.
- Ensure non-streaming Anthropic responses retain all tool_use blocks instead of discarding those after the first.

Enhancements:
- Represent streamed Anthropic tool calls as an index-keyed list on the scope, mirroring the OpenAI instrumentor’s handling of parallel tool calls.
- Comma-join multiple tool names, ids, and arguments into single-valued OTel tool attributes in a way that is byte-identical to previous behavior for the single-call case.

Tests:
- Add offline tests covering parallel Anthropic tool_use streaming, split argument deltas, interleaved text blocks, multi-call non-streaming responses, legacy single-call shape handling, and skipping empty tool entries.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/anthropic/anthropic.py
+++ b/sdk/python/src/openlit/instrumentation/anthropic/anthropic.py
@@ -61,9 +61,7 @@ def messages(
             self._output_tokens = 0
             self._cache_read_input_tokens = 0
             self._cache_creation_input_tokens = 0
-            self._tool_arguments = ""
-            self._tool_id = ""
-            self._tool_name = ""
+            self._tools = []
             self._tool_calls = None
             self._response_role = ""
             self._kwargs = kwargs
@@ -246,9 +244,7 @@ def messages_stream(
             self._output_tokens = 0
             self._cache_read_input_tokens = 0
             self._cache_creation_input_tokens = 0
-            self._tool_arguments = ""
-            self._tool_id = ""
-            self._tool_name = ""
+            self._tools = []
             self._tool_calls = None
             self._response_role = ""
             self._kwargs = kwargs
--- a/sdk/python/src/openlit/instrumentation/anthropic/async_anthropic.py
+++ b/sdk/python/src/openlit/instrumentation/anthropic/async_anthropic.py
@@ -62,9 +62,7 @@ def async_messages(
             self._output_tokens = 0
             self._cache_read_input_tokens = 0
             self._cache_creation_input_tokens = 0
-            self._tool_arguments = ""
-            self._tool_id = ""
-            self._tool_name = ""
+            self._tools = []
             self._tool_calls = None
             self._response_role = ""
             self._kwargs = kwargs
@@ -248,9 +246,7 @@ def async_messages_stream(
             self._output_tokens = 0
             self._cache_read_input_tokens = 0
             self._cache_creation_input_tokens = 0
-            self._tool_arguments = ""
-            self._tool_id = ""
-            self._tool_name = ""
+            self._tools = []
             self._tool_calls = None
             self._response_role = ""
             self._kwargs = kwargs
--- a/sdk/python/src/openlit/instrumentation/anthropic/utils.py
+++ b/sdk/python/src/openlit/instrumentation/anthropic/utils.py
@@ -444,6 +444,9 @@ def _tool_slot(scope, index):
     the openai instrumentor uses for streamed tool calls.
     """
 
+    # The four wrappers all seed `_tools = []`, but process_chunk is also driven
+    # directly (tests, and any caller building its own scope), so create the list
+    # rather than assume the constructor ran.
     tools = getattr(scope, "_tools", None)
     if tools is None:
         tools = scope._tools = []
--- a/sdk/python/src/openlit/instrumentation/anthropic/utils.py
+++ b/sdk/python/src/openlit/instrumentation/anthropic/utils.py
@@ -200,7 +200,9 @@ def build_output_messages(response_text, finish_reason, tool_calls=None):
     Args:
         response_text: Response text from model
         finish_reason: Finish reason from Anthropic (end_turn, max_tokens, stop_sequence, tool_use)
-        tool_calls: Optional tool calls dict from response
+        tool_calls: Optional tool call(s) from the response. Either a single
+            dict (legacy shape) or a list of dicts, one per parallel tool_use
+            block in the turn.
 
     Returns:
         List with single OutputMessage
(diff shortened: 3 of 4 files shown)


Kayvan-Zahiri:
Thanks for the review — both points addressed in the latest push.

**Non-streaming test now runs the real path.** You were right that it asserted on a reconstruction rather than the instrumentation. It drives `process_chat_response` with a real tracer and asserts on the exported span, so a regression in that function fails the test instead of passing it. That also let me add two assertions the old shape couldn't make: that parallel calls comma-join into the single-valued OTel tool attributes, and that a single call still produces exactly the pre-fix attribute values.

**`_tool_slot`** keeps the defensive creation, now with a comment saying why: all four wrappers seed `_tools = []`, but `process_chunk` is also driven directly by tests and by any caller assembling its own scope, so it creates the list rather than assuming the constructor ran.

8 tests passing, pylint 10.00/10 on the changed files.

Kayvan-Zahiri:
Carrying over a fix from the review on #1417, since the same issue is in this PR.

The single-valued OTel tool attributes passed `names`, `ids` and `args` through `filter(None, ...)` independently, so a call with an empty name would drop that slot from `gen_ai.tool.name` while keeping it in `.call_id` and `.args` — the nth entry of the three attributes could then describe different calls.

`_join_tool_field` now gives every call a slot in every field so the columns stay aligned, and collapses an all-empty field to `""`. The single-call case is still byte-identical to before the fix, which `test_non_streaming_single_call_attributes_are_unchanged` covers.

10 tests passing, pylint 10.00/10.

amanagarwal042:
this in favor of https://github.com/openlit/openlit/pull/1399 as it was raised earlier with same changes

Thanks @Kayvan-Zahiri for contribution, Hoping to see more PRs and issues from you soon

PR #1417, opened 2026-07-29
fix(google_ai_studio): keep every parallel function_call part

### LINKED ISSUE #1400: Bug: Gemini (google_ai_studio) instrumentor drops parallel function_call parts

Kayvan-Zahiri:
Fixes #1400. Same root cause as #1398 (PR #1416), different shape.

Gemini returns one part per function call, so parallel calls arrive as sibling `function_call` parts and can be split across streamed chunks. Both paths read `parts[0]` only:

**Non-streaming** (`process_chat_response`): `parts[0].get("function_call")` inspects the first part, so a second or third call in the same candidate was never read.

**Streaming** (`process_chunk`): that same `parts[0]` read was assigned straight into `scope._tools` on *every* chunk. A later chunk overwrote calls collected earlier, and a chunk carrying only text reset `_tools` to `None`, discarding everything.

## Fix

A shared `_function_calls(parts)` helper returns every `function_call` payload in a candidate, accepting both dict-shaped and object-shaped parts. Non-streaming uses it directly. Streaming appends to a list instead of replacing, so calls accumulate across chunks and a text-only chunk no longer clears them.

`build_output_messages` emits one `tool_call` part per call and still accepts a single dict, so the previous shape keeps working.

The single-valued OTel tool attributes comma-join across calls, matching what the `openai` and `anthropic` instrumentors do. **A single call produces byte-identical attributes to before.**

## Verification

The report's repro now reads both calls:

```
[{'name': 'get_weather', 'args': {'city': 'nyc'}},
 {'name': 'get_weather', 'args': {'city': 'sf'}}]
```

Added `tests/test_google_ai_studio_parallel_tools.py`, offline and needing no API key:

- parallel calls split across two streamed chunks accumulate rather than overwrite
- parallel calls inside a single chunk are all kept
- a text-only chunk does not clobber previously collected calls
- non-streaming reads every `function_call` part
- output messages carry one `tool_call` part per call
- object-shaped parts work, and `None` payloads are skipped
- the legacy single-dict shape still renders

```
7 passed
pylint 10.00/10 (.pylintrc, changed files)
```

## Summary by Sourcery

Handle multiple parallel Gemini function_call parts without dropping or overwriting them, and add tests to cover non-streaming and streaming scenarios.

Bug Fixes:
- Preserve all parallel function_call parts from Gemini responses in both non-streaming and streaming paths instead of only reading the first or overwriting earlier calls.
- Ensure text-only streamed chunks no longer clear previously collected function_call data.
- Maintain compatibility with legacy single-function_call inputs while correctly emitting one tool_call part per function call in output messages.
- Aggregate multiple tool call attributes into comma-joined OTel span attributes consistent with other instrumentors.

Tests:
- Add unit tests for parallel function_call handling across streamed chunks, within single chunks, for text-only chunks, for object-shaped parts, and for legacy single-dict shapes.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/google_ai_studio/utils.py
+++ b/sdk/python/src/openlit/instrumentation/google_ai_studio/utils.py
@@ -177,6 +177,21 @@ def build_input_messages(contents, system_instruction=None):
     return otel_messages
 
 
+def _join_tool_field(values):
+    """
+    Join one field across parallel tool calls, keeping positions aligned.
+
+    Every call contributes a slot, so the nth entry of gen_ai.tool.name, .call_id
+    and .args always describes the same call even when one of them is empty.
+    Filtering each field independently would shift the columns out of step.
+    A field that is empty for every call collapses to "" rather than a run of
+    separators, which keeps the common single-call case unchanged.
+    """
+
+    values = [str(value) if value else "" for value in values]
+    return ", ".join(values) if any(values) else ""
+
+
 def _function_calls(parts):
     """
     Return every ``function_call`` payload carried by a candidate's parts.
@@ -618,17 +633,17 @@ def common_chat_logic(
         # Parallel calls collapse into the single-valued OTel tool attributes
         # the same way the openai instrumentor does it: comma-joined, leaving
         # the one-call case byte-identical to before.
-        names = [call.get("name", "") for call in calls]
-        ids = [str(call.get("id", "")) for call in calls]
-        args = [str(call.get("args", "")) for call in calls]
         scope._span.set_attribute(
-            SemanticConvention.GEN_AI_TOOL_NAME, ", ".join(filter(None, names))
+            SemanticConvention.GEN_AI_TOOL_NAME,
+            _join_tool_field(call.get("name", "") for call in calls),
         )
         scope._span.set_attribute(
-            SemanticConvention.GEN_AI_TOOL_CALL_ID, ", ".join(filter(None, ids))
+            SemanticConvention.GEN_AI_TOOL_CALL_ID,
+            _join_tool_field(call.get("id", "") for call in calls),
         )
         scope._span.set_attribute(
-            SemanticConvention.GEN_AI_TOOL_ARGS, ", ".join(filter(None, args))
+            SemanticConvention.GEN_AI_TOOL_ARGS,
+            _join_tool_field(call.get("args", "") for call in calls),
         )
 
     # Span Attributes for Cost and Tokens
--- a/sdk/python/src/openlit/instrumentation/google_ai_studio/utils.py
+++ b/sdk/python/src/openlit/instrumentation/google_ai_studio/utils.py
@@ -177,6 +177,26 @@ def build_input_messages(contents, system_instruction=None):
     return otel_messages
 
 
+def _function_calls(parts):
+    """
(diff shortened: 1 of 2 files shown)


Kayvan-Zahiri:
Good catch, that's a real misalignment and it was mine — fixed in the latest push.

`names`, `ids` and `args` each went through `filter(None, ...)` independently, so a call with an empty name dropped that slot from `gen_ai.tool.name` while keeping it in `.call_id` and `.args`, leaving the nth entry of the three attributes describing different calls.

`_join_tool_field` now gives every call a slot in every field, so the columns stay in step. It also collapses a field that is empty for every call to `""` rather than a run of separators, which matters here specifically because Gemini never supplies tool ids — without that, two calls would have produced `", "` for `gen_ai.tool.call.id`.

Two tests added for both halves of that. 9 passing, pylint 10.00/10.

Worth flagging: I lifted the `filter(None, ...)` shape from the `openai` instrumentor (`sdk/python/src/openlit/instrumentation/openai/utils.py`, around lines 1016 and 1544), which has the same misalignment today. Happy to send a separate PR for that if you want it consistent — I left it out of this one to keep the diff on the reported bug. I've applied the same fix to #1416.

amanagarwal042:
Closing this in favor of https://github.com/openlit/openlit/pull/1401 as it was raised earlier with same changes

Thanks @Kayvan-Zahiri for contribution, Hoping to see more PRs and issues from you soon

Issue #1419, opened 2026-07-30
Bug: Dashboard crashes with React #130 when a seeded widget row is missing from openlit_widget (dangling openlit_board_widget refs)


amoha-enwithai:
### Component
OpenLIT (client)

### What happened?
On a fresh self-hosted install (`ghcr.io/openlit/openlit:1.24.2`), the **LLM dashboard** on `/home` (and the **Coding Agents** dashboard) fails to render with:

> Application error: a client-side exception has occurred (see the browser console for more information).
> Minified React error #130; args[]=undefined&args[]=

Root cause — two layers:

**1. Silent seed failure (the underlying cause).**
`importBoardLayout` in `src/client/src/lib/platform/manage-dashboard/board.ts` inserts widgets with:

```ts
await Promise.all(updatedWidgets.map(async (widget: any) => {
  return await createWidget(widget, databaseConfigId);
}));
```

`createWidget` catches errors and returns `{ err }` on failure. `Promise.all` therefore doesn't reject, and the subsequent `updateBoardLayout` writes `openlit_board_widget` rows referencing widget IDs that were never inserted into `openlit_widget`. On my install this happened for **7 widgets across 2 seeded dashboards**:

| Board | widget_id | Seed title |
|---|---|---|
| LLM dashboard | `e9b4e3de-0975-4082-912a-fc10ab22f401` | Top models |
| LLM dashboard | `527f0805-e039-4441-b2ec-4009d648f285` | Total Requests |
| Coding Agents | `9a1f1c20-3b27-4f88-9d51-1aaa22222221` | Sessions over time |
| Coding Agents | `9a1f1c20-3b27-4f88-9d51-1aaa66666664` | Pull requests |
| Coding Agents | `9a1f1c20-3b27-4f88-9d51-1aaa66666665` | Lines over time |
| Coding Agents | `9a1f1c20-3b27-4f88-9d51-1aaa66666666` | Top users by accepted lines |
| Coding Agents | `9a1f1c20-3b27-4f88-9d51-1aaa66666667` | Acceptance breakdown |

Confirm against your own install with:

```sql
SELECT b.title AS board, bw.widget_id
FROM openlit.openlit_board_widget bw
LEFT JOIN openlit.openlit_widget w ON w.id = bw.widget_id
JOIN openlit.openlit_board b ON b.id = bw.board_id
WHERE w.id IS NULL OR w.title = '';
```

**2. Client fails hard on dangling refs (the visible impact).**
[...]

PR #1420, opened 2026-07-30
fix(client): guard WidgetTypeIcon against undefined so one bad widget doesn't crash the dashboard (React #130)

amoha-enwithai:
## Summary
Renders `null` instead of `<undefined />` when `SUPPORTED_WIDGETS[widget.type]?.icon` is undefined. Prevents a single dangling `openlit_board_widget` row (mapping present, widget record missing) from crashing the whole dashboard with React error #130.

Fixes the client-side impact described in #1419. The underlying seed-race that produces dangling refs in the first place is a separate follow-up (call-site is `importBoardLayout` — `Promise.all(...map(createWidget))` swallows individual insert failures — noted in the issue).

## Change
`src/client/src/components/(playground)/manage-dashboard/board-creator/widgets/widget-renderer.tsx`:

```diff
 const WidgetTypeIcon = () => {
   const IconComponent =
     SUPPORTED_WIDGETS[widget.type as keyof typeof SUPPORTED_WIDGETS]?.icon;
+  if (!IconComponent) return null;
   return <IconComponent className="h-4 w-4" />;
 };
```

## Test plan
- [x] Manual repro on `ghcr.io/openlit/openlit:1.24.2` — with `openlit_board_widget` rows referencing missing widgets, `/home` throws React #130 before this patch; renders fine (missing widgets skipped, other widgets unaffected) after.
- [ ] No test coverage added — happy to add a unit test for `WidgetTypeIcon` if you'd like a specific style/location.

## Notes for reviewers
- Behavior change is limited to the icon slot for widgets whose `widget.type` isn't a key in `SUPPORTED_WIDGETS`. For every valid widget type the render path is unchanged (`IconComponent` resolves to the mapped Lucide icon).
- The widget-renderer's `WidgetComponent` `useMemo` block already returns `null` in its default case, and the render site is already guarded (`{WidgetComponent && <WidgetComponent … />}`), so an unknown widget type now degrades gracefully everywhere in this component.

## Summary by Sourcery

Make dashboard loading and imports resilient to missing widgets and partial widget creation failures.

New Features:
- Add seeded dashboard healing that recreates missing widgets and refreshes SQL configuration for existing widgets.

Bug Fixes:
- Prevent dangling board-widget references and unknown widget types from crashing dashboard rendering.
- Handle widget creation failures during dashboard imports with rollback of created widgets and board data.
- Return widget-fetch errors instead of silently treating failed fetches as missing widgets.

Enhancements:
- Propagate database configuration context through dashboard and widget operations.

Tests:
- Add coverage for seeded widget recovery, dangling mappings, widget-fetch failures, and dashboard import rollback behavior.


amanagarwal042 (dismissed):
Great fix @amoha-enwithai

amanagarwal042 (dismissed):

Issue #1421, opened 2026-07-30
Feature: add Letta instrumentation in TS SDK


AmadNaseem:
### Component

OpenLIT

### What feature would you like to see?

Add auto-instrumentation for the Letta agent platform's JavaScript SDK (`@letta-ai/letta-client`) in the TS SDK.

- New TS instrumentation directory: `sdk/typescript/src/instrumentation/letta/`

### Why do you need this?

Python already has `sdk/python/src/openlit/instrumentation/letta/`, but there's no TypeScript parity. Node services that build agents on Letta can't currently get agent, message, memory, and tool telemetry into OpenLIT.

### Additional context

- Python reference: `sdk/python/src/openlit/instrumentation/letta/`
- Convention: `.cursor/rules/js-sdk-genai-framework-instrumentation.mdc`

PR #1422, opened 2026-07-30
feat(ts-sdk): add Letta instrumentation

AmadNaseem:
**Issue number**: #1421

### Change description:

Adds Letta auto-instrumentation to the TypeScript SDK, for parity with the existing Python instrumentation (`sdk/python/src/openlit/instrumentation/letta/`).

- New `sdk/typescript/src/instrumentation/letta/` (`index.ts`, `wrapper.ts`, `utils.ts`), registered in `instrumentation/index.ts`, `InstrumentationType`, and `semantic-convention.ts`.
- Hooks `@letta-ai/letta-client` by module name via `InstrumentationNodeModuleDefinition`, with one `InstrumentationNodeModuleFile` per resource client. No static import, `moduleExports` typed `any`, resource classes duck-typed, method-presence guarded and idempotent. The package is a `devDependency` only.
- Wraps the agents, messages, blocks, tools, folders, and passages resource clients. Each method maps to the Python endpoint id so both SDKs emit the same span names, operation types, and attributes. Message operations that return a stream are traced through an async iterator that records chunk count, time to first token, and the collected messages.

A note on the JS SDK: `@letta-ai/letta-client` v1.12 is a newer generation than the client the Python instrumentation targets, so `update` maps to Python's `modify`, streaming goes through `messages.create(...).stream()`, and `folders` covers what Python calls `sources`. Resources this SDK version does not expose (`core_memory`, `context`, `groups`, `templates`) are not patched.

### How it was tested:

- `npx tsc --noEmit` and eslint both clean.
- New jest suite (`__tests__/letta-wrapper.test.ts`, 13 cases): create-agent, message (including streaming), memory, and tool spans with their attributes; token usage and cost; custom span attributes; early-break stream cleanup so the span ends once; tracing-suppression bypass; and the chat and non-chat error paths.

## Summary by Sourcery

Add Letta auto-instrumentation to the TypeScript SDK to trace @letta-ai/letta-client agents, messages, tools, folders, and passages with telemetry aligned to the Python SDK.

New Features:
- Introduce Letta instrumentation module and wrapper for @letta-ai/letta-client resource clients in the TypeScript SDK.
- Extend semantic conventions with Letta-specific operation types and attributes for agents, workflows, streaming, and model configuration.

Enhancements:
- Register Letta instrumentation in the global instrumentation registry and type union so it can be enabled like existing providers.

Build:
- Add @letta-ai/letta-client as a development dependency of the TypeScript SDK.

Tests:
- Add Jest tests covering Letta wrapper span naming, attributes, streaming behavior, error handling, and manual patching of resource client methods.


AmadNaseem:
Fixed in 4e09778: `body.messages` and `response.messages` now go through `JSON.stringify` directly instead of `String(...)`, so the input/output attributes carry the real message arrays rather than a coerced string. Added assertions that both round-trip through `JSON.parse`.

AmanAgarwal041 (approved):

Issue #870, opened 2025-09-11
[Bug]: AttributeError in langfuse + ollama chat


Roflaff:
### 🐛 What's Going Wrong?
The application throws an `AttributeError` when processing a streaming chat response. Specifically, in `common_chat_logic`, the code attempts to call `.get()` on `scope._tools`, which is a list instead of a dictionary. This causes the crash during the streaming process.

### 🕵️ Steps to Reproduce
1. Set up OpenLIT with Ollama instrumentation in an async environment.
2. Call a chat response using `async for` on the Ollama server (remote)
3. Observe the error occurring in `process_streaming_chat_response` and `common_chat_logic`.

### 🎯 What Did You Expect?
The streaming chat should process successfully without throwing an `AttributeError`, and the tool name should be read correctly from the `scope._tools` object.

### 📸 Any Screenshots?

<img width="746" height="807" alt="Image" src="https://github.com/user-attachments/assets/6e23a9ae-3e04-4363-ab98-cc4d7fddf91e" />

error: exception.stacktrace
```
[...]

PR #853, opened 2025-08-14
[Feat]: Fix un-raised error and add wrapping for Ollama `generate` method

Gwenn-LR:
### Overview:
Hi ! First of all, thanks again for your work, `openlit` saved me so many hours when I had a LLM request issue with or without knowing it.

I've realized this PR not for it to be merge as it, but to demonstrate 2 issues I had with `openlit` :
- An error from the `Ollama` API is handled [here](https://github.com/openlit/openlit/blob/9ca007497102067c17e4eb9c5423e98a38d87a69/sdk/python/src/openlit/instrumentation/ollama/ollama.py#L137-L158) but if it happens, `response` is not defined. In my project code, I was using those error to retry, with different parameters (for example when calling a model with `think=True` while the model isn't a reasoning one), but with your wrapper in this state, I can't access it anymore. Setting the line `response = wrapped(*args, **kwargs)` before the `try` helped me get the expected behavior.
- In the same project code, I use `Ollama` but not the `chat` method since I prefer `generate` because I want to be able to control exactly what is send to the LLM. However, your `Instrumentor` only handle `chat` and `embeddings` requests, that's why I've duplicated your code for `chat` and I'v modified it to get traces of  calls to `generate`. The sync implementation is the one I've used, but I've completed the async one to get a much cleaner PR, so this part of the code has not been tested.

If you prefer, I can divide this PR in 2 for those 2 aspects but I've prefered to do this one for now and get your feedback.

Have a nice day !

Openlit docker image version : 1.14.4
Openlit python SDK version : 1.35.0 

### Visuals (If applicable):

`ollama._client.Client.generate` call trace before the change : 
image

`ollama._client.Client.generate` call trace after the change : 
image

### Checklist:
- [x] PR name follows conventional commit format: `[Feat]: ...` or `[Fix]: ....`
- [x] Added visuals for changes (If applicable)
- [x] Checked OpenLIT [contribution guidelines](https://github.com/openlit/openlit/blob/main/CONTRIBUTING.md)

PR #1439, opened 2026-08-04
fix(instrumentation): handle tool_calls list in groq/ai21/together/reka span attrs (#1438)

### LINKED ISSUE #1438: Bug: groq/ai21/together/reka instrumentors crash (AttributeError) on any tool call because scope._tools is a list, not a dict
### Component
OpenLIT Python SDK

### What happened?
`process_chat_response` in the groq, ai21, together, and reka instrumentors sets `scope._tools` to the *list* of tool_calls the provider returned, then a later block calls `.get()` on it as if it were a single dict. Any response containing a tool call raises `AttributeError: 'list' object has no attribute 'get'` inside `process_chat_response`. Each wrapper's `except Exception as e: handle_exception(span, e)` catches it, so the outer call still returns the LLM response fine, but the span is silently marked ERROR and every attribute the rest of the block would have set (tokens, cost, response id, tool name/args/id) is dropped.

Affected files (HEAD `ad20b7b4c282f5c19239937f3bb485ec380e772e`):
- `sdk/python/src/openlit/instrumentation/groq/utils.py:425-436` sets `scope._tools` from `response_dict["choices"][0]["message"]["tool_calls"]` (a list, groq/ai21 both use the OpenAI-compatible schema) then does `scope._tools.get("function", {}).get("name", "")`.
- `sdk/python/src/openlit/instrumentation/ai21/utils.py:416-427` identical shape.
- `sdk/python/src/openlit/instrumentation/together/utils.py:260-270` identical shape, and one line is doubly broken: `scope._span.set_attribute(...).get('name','')` chains `.get` onto `set_attribute`'s return value, which is `None`.
- `sdk/python/src/openlit/instrumentation/reka/utils.py:368-378`: `scope._tools = first_message.get("tool_calls")` (Reka's own SDK types `tool_calls` as `List[ToolCall]`), then `.get("name", "")` on it.

This is the same bug class already reported and fixed for Ollama (#870 / PR #853), and generalized to most other providers by PR #1002's semconv refactor (cohere, azure_ai_inference, litellm, openai, mistral, langchain, gpt4all, premai, vertexai, pydo, gradient all switched to a safe list-handling pattern there). #1002 touched groq/ai21/reka's `utils.py` for unrelated renames but never touched this block, and never touched `together/utils.py` at all, so the fix never reached these four providers.

### Steps to reproduce
Calling the real, unmodified `process_chat_response` directly with a schema-accurate synthetic response reproduces it without needing a live API key:

```python
import openlit

openlit.init(disable_metrics=True, disable_events=True)
from openlit.instrumentation.groq.utils import process_chat_response

tracer = TracerProvider().get_tracer("repro")
response = {
    "id": "resp-1", "model": "test-model",
    "choices": [{
        "message": {"content": "hi", "tool_calls": [
            {"id": "call_1", "function": {"name": "get_weather", "arguments": "{}"}}
        ]},
    }],
    "usage": {"prompt_tokens": 3, "completion_tokens": 2},
}
with tracer.start_as_current_span("groq.chat") as span:
    process_chat_response(
        response=response, request_model="test-model", pricing_info={},
        server_port=443, server_address="api.example.com", environment="test",
        application_name="repro", metrics=None, start_time=0.0, span=span,
        capture_message_content=False, disable_metrics=True,
        model="test-model", tools=[{"type": "function"}],
        messages=[{"role": "user", "content": "what's the weather"}],
    )
```
Raises:
```
AttributeError: 'list' object has no attribute 'get'
```
Swapping the import for `openlit.instrumentation.{ai21,together}.utils.process_chat_response` (same response shape) and `openlit.instrumentation.reka.utils.process_chat_response` (needs `"responses": [...]` instead of `"choices"`, Reka's own response shape) raises the identical error in all three.

### Expected behavior
Same as the fix already landed for the other providers: build the tool name/id/args attributes from the list (comma-joined for multiple tool calls, matching the pattern used for ollama/cohere/openai etc.), not `.get()` a list as if it were a dict. The span should reflect a successful call with tool attributes set, not ERROR.

### Environment
- OpenLIT SDK version: 1.45.0 (HEAD ad20b7b4c282f5c19239937f3bb485ec380e772e)
- Python: 3.12

### Additional context
Reproduced in a clean `python:3.12-slim` container this session, calling the real registered `process_chat_response` for all four providers with a real `TracerProvider`/span (not mocked) after a real `openlit.init()`; all four raise the identical `AttributeError`, confirmed via `span.status`/`record_exception` being set exactly as `handle_exception` does. Did not exercise this against a live groq/ai21/together/reka API call, only the response-processing function with a synthetic but schema-accurate response.

Anai-Guo:
## What

`process_chat_response` in the **groq**, **ai21**, **together** and **reka** instrumentors sets `scope._tools` to the *list* of tool_calls the provider returns, then calls `.get()` on it as if it were a single dict:

```python
scope._tools = response_dict["choices"][0]["message"]["tool_calls"]   # a list
...
scope._tools.get("function", {}).get("name", "")                       # AttributeError
```

Any response containing a tool call raises `AttributeError: 'list' object has no attribute 'get'` inside `process_chat_response`. Each wrapper's `except Exception as e: handle_exception(span, e)` catches it, so the outer LLM call still returns fine — but **the span is silently marked ERROR and every remaining attribute (tokens, cost, response id, tool name/args/id) is dropped.**

`together/utils.py` had an extra defect on top: one line chained `.get("name", "")` onto the return value of `set_attribute()`, which is `None`.

This is the same bug class already fixed for Ollama (#870 / #853) and generalised to most providers by #1002's semconv refactor. #1002 touched groq/ai21/reka's `utils.py` for unrelated renames but never touched this block, and never touched `together/utils.py` at all, so the fix never reached these four providers.

Fixes #1438.

## Fix

Adopt the exact safe list-handling pattern #1002 introduced for `openai/utils.py` — normalise `scope._tools` to a list and join `name`/`id`/`arguments` across all tool calls (reka keeps its top-level `name`/`parameters` shape). A single dict, if one ever slips through, is still handled.

## Verification

Reproduced the crash from the issue with a schema-accurate synthetic response and confirmed the fix, old vs new, for the shared block logic:

```
OLD raises: AttributeError - 'list' object has no attribute 'get'
NEW: {'tool.name': 'get_weather, get_time', 'tool.id': 'call_1, call_2', 'tool.args': '{"city":"SF"}, {}'}
NEW reka: {'tool.name': 'lookup', 'tool.id': 'c1', 'tool.args': '{"q":"x"}'}
NEW single-dict: {'tool.name': 'get_weather', 'tool.id': 'call_1', 'tool.args': '{"city":"SF"}'}
```

## Summary by Sourcery

Normalize tool call handling in Groq, AI21, Together, and Reka Python instrumentations so spans correctly record tool metadata from list-based tool_calls responses.

Bug Fixes:
- Prevent AttributeError when processing list-based tool_calls in Groq, AI21, Together, and Reka chat instrumentation, ensuring spans are not erroneously marked as errors and tool-related attributes are preserved.

Enhancements:
- Aggregate tool names, ids, and arguments across multiple tool calls into span attributes for Groq, AI21, Together, and Reka providers, aligning their behavior with the OpenAI instrumentation.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/ai21/utils.py
+++ b/sdk/python/src/openlit/instrumentation/ai21/utils.py
@@ -414,16 +414,32 @@ def common_chat_logic(
 
     # Span Attributes for Tools
     if scope._tools:
+        tools = scope._tools if isinstance(scope._tools, list) else [scope._tools]
+
+        names, ids, args = (
+            zip(
+                *[
+                    (
+                        t.get("function", {}).get("name", ""),
+                        str(t.get("id", "")),
+                        str(t.get("function", {}).get("arguments", "")),
+                    )
+                    for t in tools
+                    if isinstance(t, dict) and t
+                ]
+            )
+            if tools
+            else ([], [], [])
+        )
+
         scope._span.set_attribute(
-            SemanticConvention.GEN_AI_TOOL_NAME,
-            scope._tools.get("function", {}).get("name", ""),
+            SemanticConvention.GEN_AI_TOOL_NAME, ", ".join(filter(None, names))
         )
         scope._span.set_attribute(
-            SemanticConvention.GEN_AI_TOOL_CALL_ID, str(scope._tools.get("id", ""))
+            SemanticConvention.GEN_AI_TOOL_CALL_ID, ", ".join(filter(None, ids))
         )
         scope._span.set_attribute(
-            SemanticConvention.GEN_AI_TOOL_ARGS,
-            str(scope._tools.get("function", {}).get("arguments", "")),
+            SemanticConvention.GEN_AI_TOOL_ARGS, ", ".join(filter(None, args))
         )
 
     # Span Attributes for Cost and Tokens
--- a/sdk/python/src/openlit/instrumentation/ai21/utils.py
+++ b/sdk/python/src/openlit/instrumentation/ai21/utils.py
@@ -416,21 +416,16 @@ def common_chat_logic(
     if scope._tools:
         tools = scope._tools if isinstance(scope._tools, list) else [scope._tools]
 
-        names, ids, args = (
-            zip(
-                *[
-                    (
-                        t.get("function", {}).get("name", ""),
-                        str(t.get("id", "")),
-                        str(t.get("function", {}).get("arguments", "")),
-                    )
-                    for t in tools
-                    if isinstance(t, dict) and t
-                ]
+        tool_rows = [
+            (
+                t.get("function", {}).get("name", ""),
+                str(t.get("id", "")),
+                str(t.get("function", {}).get("arguments", "")),
             )
-            if tools
-            else ([], [], [])
-        )
+            for t in tools
+            if isinstance(t, dict) and t
+        ]
+        names, ids, args = zip(*tool_rows) if tool_rows else ([], [], [])
 
         scope._span.set_attribute(
             SemanticConvention.GEN_AI_TOOL_NAME, ", ".join(filter(None, names))
--- a/sdk/python/src/openlit/instrumentation/groq/utils.py
+++ b/sdk/python/src/openlit/instrumentation/groq/utils.py
@@ -425,21 +425,16 @@ def common_chat_logic(
     if scope._tools:
         tools = scope._tools if isinstance(scope._tools, list) else [scope._tools]
 
-        names, ids, args = (
-            zip(
-                *[
-                    (
-                        t.get("function", {}).get("name", ""),
-                        str(t.get("id", "")),
-                        str(t.get("function", {}).get("arguments", "")),
-                    )
-                    for t in tools
-                    if isinstance(t, dict) and t
-                ]
+        tool_rows = [
+            (
+                t.get("function", {}).get("name", ""),
+                str(t.get("id", "")),
+                str(t.get("function", {}).get("arguments", "")),
             )
-            if tools
-            else ([], [], [])
-        )
+            for t in tools
+            if isinstance(t, dict) and t
+        ]
+        names, ids, args = zip(*tool_rows) if tool_rows else ([], [], [])
 
         scope._span.set_attribute(
             SemanticConvention.GEN_AI_TOOL_NAME, ", ".join(filter(None, names))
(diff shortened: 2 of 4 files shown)


AmanAgarwal041:
Findings on this PR:

- ✅ Fixes the exact crash from the issue: a single tool call (or multiple parallel tool calls) in `groq`/`ai21`/`together`/`reka` no longer raises `AttributeError`, and the span attributes (`gen_ai.tool.name`, `gen_ai.tool.call.id`, `gen_ai.tool.args`) are set correctly for all 4 providers.
- ⚠️ There's an edge case this PR doesn't cover: if `scope._tools` is a non-empty list but every entry fails the `isinstance(t, dict) and t` filter (e.g. an unexpected `None`/non-dict entry in `tool_calls`), the `if tools else ([], [], [])` guard checks the truthiness of the *original* list, not the filtered one. That means `zip(*[])` (empty list comprehension) gets unpacked into 3 variables, which raises `ValueError: not enough values to unpack (expected 3, got 0)` instead of gracefully falling back to empty attributes. I reproduced this directly against this branch:

  ```python
  response["choices"][0]["message"]["tool_calls"] = [None, "garbage-string"]
  # -> ValueError: not enough values to unpack (expected 3, got 0)
  #    at the "names, ids, args = (...)" line in common_chat_logic

Anai-Guo:
Thanks @AmanAgarwal041 — good catch. Fixed in the latest commit.

The root cause was exactly as you described: the `if tools else (...)` guard checked the truthiness of the *original* `tools` list, not the filtered comprehension. So `tool_calls = [None, "garbage-string"]` passed the guard (original list is truthy) but the comprehension produced `[]`, and `zip(*[])` unpacked into 3 variables → `ValueError`.

The fix materializes the filtered rows first and guards on **those**:

```python
tool_rows = [
    (
        t.get("function", {}).get("name", ""),
        str(t.get("id", "")),
        str(t.get("function", {}).get("arguments", "")),
    )
    for t in tools
    if isinstance(t, dict) and t
]
names, ids, args = zip(*tool_rows) if tool_rows else ([], [], [])
```

Applied to all four instrumentations (groq/ai21/together, and reka with its `name`/`parameters` shape). Verified against your repro:

- `tool_calls = [None, "garbage-string"]` → attributes fall back to empty strings, no exception
- single / multiple valid tool calls → unchanged behavior

amanagarwal042:
Closing this in favor of #1452. Really appreciate the time spent on this and looking forward to more future work on OpenLIT

PR #1441, opened 2026-08-04
fix(go): include Anthropic prompt cache tokens in streaming cost

### LINKED ISSUE #1440: Go SDK: streamed Anthropic calls drop cache tokens from total_tokens and from cost

```go
}
```

```go
}
```

Then at line 215:

```go

```

Solaris-star:
## Summary

Anthropic streaming responses expose `cache_read_input_tokens` and `cache_creation_input_tokens` in the `message_start` usage block. The Go SDK was parsing the fields but dropping them from the final span usage and cost calculation.

That caused Anthropic prompt-cache usage to be under-reported and costs to be too low for streaming requests.

## Changes

- Preserve cache-read and cache-creation token counts through the streaming path.
- Include both cache token classes in `gen_ai.usage.total_tokens` and the existing cache detail attributes.
- Include cache pricing in streaming cost calculation.
- Extend custom and fetched pricing data with cache rates.
- Preserve the existing `CalculateCost` API for callers that do not provide cache counts.
- Add pricing and end-to-end streaming regression tests.

## Verification

- `go test ./...`
- `go vet ./...`
- `git diff --check`

Fixes #1440

## Summary by Sourcery

Include Anthropic prompt cache usage in Go SDK streaming telemetry and cost calculation.

Bug Fixes:
- Ensure Anthropic streaming spans include cache read and cache creation token counts in usage and cost attributes.

Enhancements:
- Extend pricing models, configuration, and helpers to support cache read and cache creation token pricing, including chat pricing sources.
- Unify usage attribute handling so total token counts and cache-related attributes are consistently recorded for both standard and streaming Anthropic responses.

Tests:
- Add unit tests for pricing helpers to cover cache-aware cost calculation and fetched chat cache rates.
- Add an Anthropic streaming regression test to validate prompt cache token usage and cost reporting.

### DIFF
--- a/sdk/go/config.go
+++ b/sdk/go/config.go
@@ -97,8 +97,10 @@ type Config struct {
 
 // ModelPricing represents pricing information for a model
 type ModelPricing struct {
-	InputCostPerToken  float64
-	OutputCostPerToken float64
+	InputCostPerToken         float64
+	OutputCostPerToken        float64
+	CacheReadCostPerToken     float64
+	CacheCreationCostPerToken float64
 }
 
 // setDefaults sets default values for the configuration
--- a/sdk/go/helpers/pricing.go
+++ b/sdk/go/helpers/pricing.go
@@ -12,27 +12,37 @@ import (
 
 // PricingInfo holds pricing information for a model
 type PricingInfo struct {
-	InputCostPerToken  float64 `json:"input_cost_per_token"`
-	OutputCostPerToken float64 `json:"output_cost_per_token"`
+	InputCostPerToken         float64 `json:"input_cost_per_token"`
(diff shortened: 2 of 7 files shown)


Solaris-star:
Addressed the Sourcery review points in 4705980:

- cost assertions now use an epsilon comparison;
- added an unknown-model cost test confirming the fallback is zero;
- extended the pricing fixture and assertions to cover legacy `data` cache rates as well as `chat` cache rates.

`go test ./...`, `go vet ./...`, and `git diff --check` pass.

Solaris-star:
Addressed the remaining Sourcery test note in `c748fcb`: the end-to-end streaming cost assertion now uses an epsilon tolerance. `go test ./...`, `go vet ./...`, and `git diff --check` pass.

Issue #1427, opened 2026-08-02
fix(ts-sdk): PII redaction leaves part of the secret in the output when patterns overlap


pacocartones:
**Component:** OpenLIT TS SDK (`sdk/typescript/src/guard/pii.ts`)

### What happens

When two PII/secret patterns produce overlapping matches on the same text, `REDACT` corrupts the output — and in the nesting case, leaves the tail of a secret value unredacted. The replacement loop sorts matches by descending `start` and slices each span out of the original text **without merging overlaps first** (`pii.ts:91-99`). Because every offset refers to the original string, replacing an inner span first shifts the `end` of the outer span, so the outer replacement stops early and the tail of the value survives.

### Steps to reproduce

Run `new PII({ action: "redact" })` from current `main` (`176e9608`) on these inputs (all executed literally here, outputs byte-for-byte):

| Input | Output on `main` | Problem |
|---|---|---|
| `password: a@b.co:hunter2SuperSecret` | `[REDACTED:env-secret]uperSecret` | **partial leak**: 10 chars of the value's tail survive |
| `secret: 1.2.3.4-prod-fallback-value` | `[REDACTED:env-secret]ck-value` | **partial leak**: 8 chars survive |
| `token=a@b.com,tail` | `[REDACTED:env-secret]ail],tail` | leak + corruption (4 chars + orphan `]`) |
| `my password: hunter2@corp.io then continue` | `my [REDACTED:env-secret]] then continue` | **format corruption only** — the email IS fully covered, just an orphan `]` |
| `Key: sk-ant-abcdefghij1234567890abcde` | `Key: [REDACTED:anthropic-api-key]de` | 2 chars of the key survive — this input comes from the repo's own test suite |

For contrast, disjoint spans work correctly (`Email: a@b.com and SSN: 123-45-6789` → both placeholders, no artifacts), and adjacent-but-not-overlapping secrets are fine too.

### Expected behavior

Overlapping matches are merged (or otherwise handled) before replacement, so a nested secret is fully covered by a single placeholder and no orphan brackets survive. Note that fixing this changes the output of 3 existing tests — I would update those expectations in the same PR.

### Scope note

This is redaction hardening / correctness work, not a remotely exploitable vulnerability: the redactor is opt-in (`setupAutoGuards(guards)`), and the surviving fragments are tails of values rather than recoverable secrets in most cases. The Python SDK's `pii.py` appears to use the same replacement pattern — mentioned for completeness, not verified here.

[...]

PR #1444, opened 2026-08-09
fix(python-sdk): merge overlapping PII spans before redacting

pacocartones:
**Issue number**: #1427

### Change description:

The Python SDK has the same overlapping-span redaction bug reported for the TypeScript SDK in #1427 (TS side addressed by #1429). The two SDKs are meant to stay in sync, so this is the Python parity fix.

When two patterns match overlapping spans of the same text, every span's offsets refer to the original string. Replacing them one at a time in descending-start order lets an earlier replacement resize the string out from under a later one — the later slice then stops short and **leaves the tail of the secret in the output**. Reproduced on current `main`:

```python
from openlit.guard.pii import PII
guard = PII(action="redact")
guard.evaluate("token=1.1.1.1secretos").transformed_text
# main:  '[REDACTED:env-secret]secretos'   <- tail of the secret left behind
# fixed: '[REDACTED:env-secret]'
```

The fix coalesces overlapping spans first (sort by start, merge when `next.start < current.end`), then applies the descending slice-replace over disjoint intervals only — the same approach as #1429 on the TS side, so both SDKs keep identical redaction semantics. Adds two regression tests mirroring the existing test style.

### Verification

Run against this branch (`sdk/python`, Python 3.12, Windows):

```
$ pytest tests/test_guard_pii.py -q          # with the fix
20 passed

$ pytest tests/test_guard_pii.py tests/test_guard_integration.py \
         tests/test_guard_pipeline.py tests/test_guard_prompt_injection.py -q
63 passed
```

Red/green check (fix reverted locally, regression tests kept): both new tests fail on `main`'s implementation with the literal symptom `[REDACTED:env-secret]secretos`; they pass with the fix.

### Checklist

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/openlit/openlit/pulls) for the same update/change? (#1429 covers the TypeScript SDK only; this is the Python parity fix for the same issue)
* [x] Changes are documented

### Acknowledgment

## Summary by Sourcery

Merge overlapping PII matches before redaction to ensure complete and consistent secret removal.

Bug Fixes:
- Prevent overlapping PII matches from leaving portions of secrets unredacted.
- Preserve separate redactions for disjoint and adjacent PII spans.

Tests:
- Add regression coverage for overlapping, disjoint, and adjacent PII redaction spans.


pacocartones:
Good call — addressed in `75c6292`: `test_adjacent_spans_still_redact_separately` uses a Twilio key immediately followed by a fine-grained GitHub token (`SK" + "a"*32 + "github_pat_" + "b"*22`), and asserts both placeholders survive (`[REDACTED:twilio-api-key][REDACTED:github-fine-grained]`). Verified locally before pushing: touching spans (where `next.start == prev.end`) are not merged, exactly because the merge condition is strict `<`.

amanagarwal042 (approved):

amanagarwal042 (approved):

PR #1450, opened 2026-08-11
fix(python-sdk): relax langchain dependency constraint to >=0.3.15

### LINKED ISSUE #949: The latest openlit version is depending on an an old langchain version
According to the pyproject.toml it depends on langchain >0.3.15 and <0.4.0 however the latest langchain version is already at 1.1.3.

So I can't use openlit in my project which unfortunately cannot be downgraded.

VedantMadane:
### Summary

Fixes #949

In sdk/python/pyproject.toml, the langchain dependency was constrained to ^0.3.15 (>=0.3.15, <0.4.0), causing installation and resolution conflicts for projects using newer versions of langchain.

This PR relaxes the dependency constraint to langchain =

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -45,6 +45,7 @@ opentelemetry-instrumentation-urllib3 = ">=0.52b0,<1.0.0"
 opentelemetry-instrumentation-system-metrics = {version = ">=0.52b0,<1.0.0", optional = true}
 openai = ">=1.92.0,<3.0.0"
 anthropic = ">=0.42.0,<1.0.0"
+langchain = ">=0.3.15"
 
 [tool.poetry.scripts]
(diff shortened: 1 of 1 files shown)


VedantMadane:
Automated fix: resolved merge conflicts and rebased onto main.

amanagarwal042:
Thanks for the contribution. The original intent (don't pin users to langchain 0.3.x) is right, but this PR no longer matches current main.
#949 was already fixed by #1063, which removed langchain from sdk/python/pyproject.toml entirely. OpenLIT instruments langchain-core when it is present; it should not install langchain for every user.

This reintroduces a hard dependency we deliberately removed. It also fights python = "^3.9.0" (langchain 1.x needs 3.10+) and is unbounded, unlike openai/anthropic.
So Im closing this. No further change is needed on main for this issue.

PR #1451, opened 2026-08-12
fix(sdk): resolve hyphen/underscore instrumentor name spellings in disabled_instrumentors

Anai-Guo:
## What

`openlit.init(disabled_instrumentors=[...])` matches names against the canonical keys of `MODULE_NAME_MAP`, and those keys are inconsistent about their separator — `llama_index` and `pydantic_ai` sit next to `openai-agents`, `google-ai-studio` and `aiohttp-client`.

`INSTRUMENTOR_ALIASES` exists to absorb exactly that ("Common aliases so users can pass intuitive names…"), but it only covers *one* direction: the underscore spelling of a hyphenated key. The reverse — the hyphenated spelling of an underscore key — has no entry, so five spellings resolve to nothing:

| passed in | should resolve to | today |
|---|---|---|
| `llama-index` | `llama_index` | unresolved |
| `llamaindex` | `llama_index` | unresolved |
| `pydantic-ai` | `pydantic_ai` | unresolved |
| `reka_api` | `reka-api` | unresolved |
| `aiohttp_client` | `aiohttp-client` | unresolved |

`llama-index` and `pydantic-ai` are the distribution names the instrumentations themselves declare (`_instruments = ("llama-index >= 0.10.0",)`, `("pydantic-ai >= 0.2.17",)`), and `llamaindex` is the name of the package directory — so all three are spellings the repo already uses elsewhere.

## Why it matters

An unresolved name is not an error. `__init__.py` puts it on `invalid_instrumentors`, logs a warning and moves on, so the instrumentation the user asked to turn off keeps emitting spans. The "Did you mean" suggester does not help either: it is a substring match, and `"llama-index"` is not a substring of `"llama_index"` (nor the reverse), so these names fall through to the generic message with no hint at the fix.

Running the validation block from `openlit/__init__.py` verbatim against `disabled_instrumentors=["llama-index", "llamaindex", "pydantic-ai", "reka_api"]`:

```
--- before
  resolved: ['llama-index', 'llamaindex', 'pydantic-ai', 'reka_api']
  effectively disabled: []
  WARNING - Invalid instrumentor name detected and ignored: 'llama-index'
  WARNING - Invalid instrumentor name detected and ignored: 'llamaindex'
  WARNING - Invalid instrumentor name detected and ignored: 'pydantic-ai'
  WARNING - Invalid instrumentor name detected and ignored: 'reka_api'

--- after
  resolved: ['llama_index', 'llama_index', 'pydantic_ai', 'reka-api']
  effectively disabled: ['llama_index', 'llama_index', 'pydantic_ai', 'reka-api']
```

## Change

Five entries added to `INSTRUMENTOR_ALIASES`, in the existing style and next to their siblings. No lookup semantics change — the same table, filled in the other direction.

`sdk/python/tests/test_instrumentor_names.py` is new and is what keeps this from drifting back: `test_both_separator_spellings_resolve` walks every key in `MODULE_NAME_MAP` and asserts both spellings normalize onto a real instrumentor, so a future key with a new separator mix fails the suite instead of silently ignoring a user's opt-out. The rest of the file pins the invariants the two registries already satisfy (aliases point at real keys, aliases do not shadow keys, `MODULE_NAME_MAP` and `INSTRUMENTOR_MAP` agree).

## Verification

Against `main` the new test file is 4 failed / 4 passed; with this commit it is 8 passed:

```
=== before
FAILED test_instrumentor_names.py::test_both_separator_spellings_resolve
FAILED test_instrumentor_names.py::test_distribution_names_resolve
FAILED test_instrumentor_names.py::test_module_directory_names_resolve
FAILED test_instrumentor_names.py::test_normalization_is_case_insensitive
4 failed, 4 passed

=== after
8 passed
```

`pylint` with `sdk/python/.pylintrc` rates both changed files 10.00/10.

Happy to fold the aliases into a generic separator-swap fallback inside `normalize_instrumentor_name` instead, if you would rather not grow the table — I kept it as explicit entries because that is the mechanism already in place.

## Summary by Sourcery

Ensure common instrumentor name spellings reliably map to their canonical keys when processing disabled instrumentors.

Bug Fixes:
- Resolve hyphenated, underscored, and unseparated instrumentor name variants so disabled instrumentors are correctly recognized.

Enhancements:
- Add coverage for instrumentor aliases, canonical registry consistency, distribution and module naming variants, case-insensitive normalization, and unknown-name behavior.

Tests:
- Add unit tests that prevent unresolved instrumentor spellings and registry inconsistencies from regressing.


amanagarwal042 (approved):
Looks Great!

PR #1452, opened 2026-08-12
fix(groq,ai21,together,reka): handle tool_calls list in span attributes

### LINKED ISSUE #1438: Bug: groq/ai21/together/reka instrumentors crash (AttributeError) on any tool call because scope._tools is a list, not a dict

Aftabbs:
## Summary

Fixes #1438.

Four instrumentors — `groq`, `ai21`, `together`, and `reka` — crashed
with `AttributeError: 'list' object has no attribute 'get'` on any
response that contained tool calls. The error was silently swallowed by
the outer `except Exception` handler in `common_chat_logic`, so the LLM
call appeared to succeed — but the span was recorded as `ERROR` and
every tool/token/cost attribute was dropped.

**Root cause:** `scope._tools` is assigned the full list returned by
`response["choices"][0]["message"]["tool_calls"]`, but the code then
calls `.get()` on it as if it were a single dict.

The `together` provider had a second bug: the tool-name was accidentally
chained onto the `None` return value of `span.set_attribute` via
`.get("name", "")`, causing `AttributeError: 'NoneType' object has no
attribute 'get'` unconditionally even for a single-item list.

## Fix

Normalise `scope._tools` to a list and zip-extract names, IDs, and
arguments in one pass, then join with `", "` — identical to the pattern
applied to the `cohere`, `openai`, and `mistral` providers in #1002.

`reka` uses different field names (`name`/`id`/`parameters` instead of
`function.name`/`id`/`function.arguments`) and can receive either plain
dicts or SDK objects, so it uses `getattr` fallbacks accordingly.

## Files changed

| File | Change |
|------|--------|
| `sdk/python/src/openlit/instrumentation/groq/utils.py` | fix list-handling in `common_chat_logic` |
| `sdk/python/src/openlit/instrumentation/ai21/utils.py` | same fix |
| `sdk/python/src/openlit/instrumentation/together/utils.py` | same fix + remove chained `.get()` on `set_attribute` return |
| `sdk/python/src/openlit/instrumentation/reka/utils.py` | same fix with Reka-specific field names |
| `sdk/python/tests/test_tool_call_list_attrs.py` | 9 regression tests (new file) |

## Test plan

```
pytest sdk/python/tests/test_tool_call_list_attrs.py -v
```

Expected:
```
PASSED test_groq_single_tool_call_does_not_raise
PASSED test_groq_parallel_tool_calls_are_joined
PASSED test_groq_no_tool_calls_skips_tool_attrs
PASSED test_ai21_single_tool_call_does_not_raise
PASSED test_ai21_parallel_tool_calls_are_joined
PASSED test_together_single_tool_call_does_not_raise
PASSED test_together_parallel_tool_calls_are_joined
PASSED test_reka_single_tool_call_does_not_raise
PASSED test_reka_parallel_tool_calls_are_joined
9 passed
```

All 9 pass locally against the fixed code; all 9 fail against the original code.

## Summary by Sourcery

Handle tool-call lists consistently across Groq, AI21, Together, and Reka instrumentation.

Bug Fixes:
- Prevent Groq, AI21, Together, and Reka instrumentors from failing when responses contain tool-call lists, preserving tool, token, and cost span attributes.
- Capture names, IDs, and arguments from multiple tool calls as comma-separated span attributes across supported providers.

Tests:
- Add regression coverage for single, parallel, missing, and invalid tool-call responses across the affected instrumentors.


AmanAgarwal041:
I verified this PR by checking out the branch and calling the real `process_chat_response` for all four providers with schema-accurate synthetic responses (no live API keys needed), and also ran the added test suite:

- ✅ Fixes the exact crash from the issue for `groq`/`ai21`/`together`/`reka` — single and parallel tool calls no longer raise `AttributeError`, span attributes are set correctly, and the `together`-specific `.get()` chained onto `set_attribute`'s `None` return value is also fixed.
- ✅ Handles the edge case that #1439 misses: when `scope._tools` is a non-empty list but every entry fails the `isinstance(t, dict) and t` filter (e.g. stray `None`/non-dict entries), this PR checks `if filtered_tools` (the post-filter list) rather than the raw `tools` list, so it falls back to `(), (), ()` instead of crashing with `ValueError: not enough values to unpack`. I confirmed #1439's version does crash on this exact input against its branch, while this PR returns `StatusCode.OK` with empty tool attributes.
- ✅ All 10 tests in `sdk/python/tests/test_tool_call_list_attrs.py` (including `test_groq_all_tools_filtered_out_does_not_raise`, which exercises the edge case above) pass against this branch.

Nothing major missing from a correctness standpoint — this is the more robust of the two fixes. A couple of minor nits if you want to polish it further:
- The new `filtered_tools`/`zip` block is duplicated near-verbatim across `groq`, `ai21`, and `together` `utils.py`; could be worth factoring into a small shared helper (e.g. in a common instrumentation utils module) to avoid drift if this pattern needs to change again later, though that's a nice-to-have, not a blocker.
- Might be worth double-checking `black`/`pylint` formatting on the new test file and the modified `utils.py` blocks match repo conventions before merge.

amanagarwal042 (approved):

PR #1462, opened 2026-08-16
fix(go/anthropic): set prompt-cache token attributes on streamed spans (#1440)

### LINKED ISSUE #1440: Go SDK: streamed Anthropic calls drop cache tokens from total_tokens and from cost

Anai-Guo:
Fixes #1440.

## What's wrong

`Usage` carries three prompt-token counters and the streaming path reads one of them.

`sdk/go/instrumentation/anthropic/streaming.go`, at `message_start`:

```go
}
```

`CacheCreationInputTokens` and `CacheReadInputTokens` are declared right beside `InputTokens` in `types.go:72-73` and are already unmarshalled from the event — they are simply never read, so `grep -i cache streaming.go` finds nothing.

The non-streaming path does record them (`instrumentor.go:202-207`):

```go
if resp.Usage.CacheCreationInputTokens > 0 {
    semconv.SetIntAttribute(span, semconv.GenAIUsagePromptTokensDetailsCacheWrite, resp.Usage.CacheCreationInputTokens)
}
}
```

So the same Anthropic call emits `gen_ai.usage.prompt_tokens_details.cache_read` / `.cache_write` when it is not streamed, and emits neither when it is.

## The fix

Capture both counters at `message_start` and set the same two attributes the non-streaming path sets, with the same `> 0` guards. +11 lines, one file.

## Scope: no already-published number moves

Issue #1440 also asks whether `total_tokens` and `cost` should grow by the cached amount. This PR deliberately does **not** touch either, because both paths already agree on them:

| | non-streaming (`instrumentor.go`) | streaming (`streaming.go`) |
|---|---|---|
| `total_tokens` | `InputTokens + OutputTokens` (L201) | `inputTokens + outputTokens` (L218) |
| `cost` | `CalculateGlobalCost(model, input, output)` (L102) | `CalculateGlobalCost(model, input, output)` (L220) |

Cache tokens are excluded from the total and from cost in **both** paths, so raising them in the streamed path alone would introduce a new inconsistency rather than remove one. Whether cache tokens should count toward `total_tokens`/`cost` is a cross-path pricing decision that affects every instrumentation and every existing dashboard — happy to follow up on it separately if you want that changed, but it seemed wrong to fold a published-number change into a missing-attribute fix.

The gap this PR closes is unambiguous: two attributes present on one path and absent on the other, from data that is already parsed.

## Notes

- An earlier, broader attempt at #1440 (#1441, which also reworked pricing) was closed by its author; this is the narrow parity-only subset of it.
- No test asserts span attributes anywhere in `sdk/go` today (the instrumentation tests run with `DisableTracing: true` and assert stream content only), so this follows the existing convention rather than introducing a span-recorder harness for one attribute pair. Verified statically instead: both constants exist in `semconv/genai.go:115,118` and `SetIntAttribute(span, string, int)` matches the `int` counters.

## Summary by Sourcery

Align streamed Anthropic usage instrumentation with the non-streaming path by preserving prompt-cache attributes without changing total-token or cost calculations.

Bug Fixes:
- Record prompt-cache read and write token attributes for streamed Anthropic responses, matching non-streaming instrumentation.
- Ensure streaming spans are emitted when usage contains only cache-token counts.

Tests:
- Add streaming instrumentation tests covering cache-token attributes, cache-only usage, and unchanged total-token accounting.


Anai-Guo:
Update on verification — the PR description says this was verified statically only. I've since been able to run the Go toolchain, so here are actual results on this branch (go1.26.5, `sdk/go`):

```
go build ./...   OK
go vet ./...     OK
gofmt -l         clean
go test ./...    ok  sdk/go, anthropic, openai, vllm   (no test files: helpers, semconv)
```

I also ran a throwaway probe (**not** included in the diff, since `sdk/go` has no span-recorder harness) that installs a `tracetest.SpanRecorder` as the global tracer provider and streams a `message_start` carrying both cache counters:

```json
"usage":{"input_tokens":10,"output_tokens":0,
         "cache_creation_input_tokens":1234,"cache_read_input_tokens":5678}
```

| | `cache_read` | `cache_write` | `total_tokens` |
|---|---|---|---|
| upstream `main` | *absent* | *absent* | 12 |
| this branch | 5678 | 1234 | 12 |

So the two attributes go from absent to correct, and `total_tokens` stays 12 (`input 10 + output 2`) on both sides — confirming empirically what the description argues: no already-published number moves. Cost is likewise untouched.

Happy to fold the probe in as a real test if you'd like a span-recorder harness in `sdk/go`; I left it out to keep this a one-file change consistent with the existing tests, which all run with `DisableTracing: true`.

amanagarwal042 (approved):

PR #1475, opened 2026-08-18
fix(go/anthropic): propagate cache tokens to streaming spans

### LINKED ISSUE #1440: Go SDK: streamed Anthropic calls drop cache tokens from total_tokens and from cost

Aftabbs:
## Problem

As described in #1440, the Go SDK's streaming path reads only `input_tokens` from the `message_start` SSE event and discards the two cache token fields that Anthropic reports beside it:

```go
// before — only InputTokens was read
}
```

`CacheReadInputTokens` and `CacheCreationInputTokens` are declared right next to `InputTokens` in `types.go`, so the data arrives in the struct and is silently dropped.  This produces three observable gaps in the resulting span:

| Attribute | Streaming (before) | Non-streaming |
|-----------|-------------------|---------------|
| `gen_ai.usage.prompt_tokens_details.cache_read` | absent | set |
| `gen_ai.usage.prompt_tokens_details.cache_write` | absent | set |
| `gen_ai.usage.total_tokens` | short by cached amount | also short (same underlying issue) |

## Fix

Three minimal changes to `streaming.go`:

1. **Declare** `cacheReadInputTokens` and `cacheCreationInputTokens` alongside the existing token variables.
2. **Read** both fields from `event.Message.Usage` in the `message_start` case.
3. **Set** the two cache span attributes behind `> 0` guards, mirroring the non-streaming path in `instrumentor.go` exactly; include both cache counts in `gen_ai.usage.total_tokens`.

### Diff in words

```
- var inputTokens, outputTokens int
+ var inputTokens, outputTokens, cacheReadInputTokens, cacheCreationInputTokens int

+     cacheReadInputTokens = event.Message.Usage.CacheReadInputTokens
+     cacheCreationInputTokens = event.Message.Usage.CacheCreationInputTokens
  }

- semconv.SetIntAttribute(span, semconv.GenAIUsageTotalTokens, inputTokens+outputTokens)
+ semconv.SetIntAttribute(span, semconv.GenAIUsageTotalTokens, inputTokens+outputTokens+cacheReadInputTokens+cacheCreationInputTokens)
+ if cacheCreationInputTokens > 0 {
+     semconv.SetIntAttribute(span, semconv.GenAIUsagePromptTokensDetailsCacheWrite, cacheCreationInputTokens)
+ }
+ if cacheReadInputTokens > 0 {
+     semconv.SetIntAttribute(span, semconv.GenAIUsagePromptTokensDetailsCacheRead, cacheReadInputTokens)
+ }
```

## Note on cost

`CalculateGlobalCost` accepts only `inputTokens, outputTokens`; Anthropic prices cache reads and writes at different rates than regular input, so accurate per-call cache cost would require extending that helper.  That is left for a follow-up to keep this PR small — the total-token count already improves from this change and the cache attributes now appear on streaming spans.

## Test

`TestCreateMessage_Streaming_WithCacheTokens` exercises the streaming path with a `message_start` event that carries `cache_read_input_tokens: 200`, confirming the stream drains without error and content is accumulated correctly.

Closes #1440

## Summary by Sourcery

Propagate cached token usage from Anthropic streaming responses so telemetry matches non-streaming spans.

Bug Fixes:
- Propagate Anthropic cache-read and cache-creation token usage to streaming span attributes and include cached tokens in total usage counts.

Tests:
- Add streaming coverage for cache-token responses, including cache-only usage scenarios.

### DIFF
--- a/sdk/go/instrumentation/anthropic/anthropic_test.go
+++ b/sdk/go/instrumentation/anthropic/anthropic_test.go
@@ -554,3 +554,56 @@ func TestCreateMessage_Streaming_WithCacheTokens(t *testing.T) {
 		t.Errorf("want accumulated content %q, got %q", "Cached", accumulated)
 	}
 }
+
+// TestCreateMessage_Streaming_CacheOnlySpan verifies that span attributes are
+// recorded even when input_tokens=0 and output_tokens=0 but cache_read tokens
+// are present.  Previously the guard `if inputTokens > 0 || outputTokens > 0`
+// skipped the entire usage block, dropping cache-read attributes.
+func TestCreateMessage_Streaming_CacheOnlySpan(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.Header().Set("Cache-Control", "no-cache")
+		w.WriteHeader(http.StatusOK)
+
+		events := []struct{ event, data string }{
+			// input_tokens=0, output_tokens=0, but cache_read_input_tokens=500.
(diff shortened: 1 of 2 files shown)


amanagarwal042:
closing this in favor of #1462

Issue #1495, opened 2026-08-29
Bug: postflight guards silently skipped for Mistral's async chat method


aniketwaghh:
### Component

OpenLIT Python SDK

### What happened?

`setup_auto_guards` picks the wrapper for each entry in `GUARDED_METHODS` with
`is_async = "Async" in class_method`. Every provider in that table follows the
`AsyncCompletions` / `AsyncMessages` / `AsyncClientV2` naming except Mistral, which spells its
coroutine method `Chat.complete_async`. So that one entry gets `_make_sync_guard_wrapper`.

The sync wrapper calls `wrapped(...)` without awaiting, hands the resulting coroutine object to
`_apply_postflight`, and the extractor finds no `choices` / `content` / `text` on a coroutine and
returns `""`, which short-circuits the pipeline. The coroutine is returned unchanged, the caller
awaits it and gets the real response, so nothing errors and preflight still works. The response
half of the pipeline is simply never run.

Concretely, with `openlit.init(guards=[openlit.PII(action="redact")])`, a secret in the model's
answer is redacted for `client.chat.complete()` and passes through untouched for
`client.chat.complete_async()`. The same applies to `action="deny"`: an async Mistral call cannot
be blocked on its output.

### Steps to reproduce

No API key needed: the Mistral SDK's HTTP client is swapped for an `httpx.MockTransport`.

```python
import asyncio

import httpx
import openlit
from mistralai import Mistral

LEAK = "Reach me at victim@example.com"


def handler(request):
    return httpx.Response(
        200,
        json={
            "id": "cmpl-1",
            "object": "chat.completion",
            "created": 0,
            "model": "mistral-small-latest",
            "choices": [
                {
                    "index": 0,
                    "message": {"role": "assistant", "content": LEAK},
                    "finish_reason": "stop",
                }
            ],
            "usage": {"prompt_tokens": 5, "completion_tokens": 5, "total_tokens": 10},
        },
    )


openlit.init(guards=[openlit.PII(action="redact")], disable_metrics=True, disable_batch=True)

transport = httpx.MockTransport(handler)
client = Mistral(
    api_key="test-key",
    client=httpx.Client(transport=transport),
    async_client=httpx.AsyncClient(transport=transport),
)
messages = [{"role": "user", "content": "hello"}]

sync_response = client.chat.complete(model="mistral-small-latest", messages=messages)
async_response = asyncio.run(
    client.chat.complete_async(model="mistral-small-latest", messages=messages)
)

[...]

PR #1496, opened 2026-08-29
fix(python-sdk): run postflight guards on Mistral async chat calls

aniketwaghh:
> [!IMPORTANT]
> 2. PR title follows Conventional Commit format, for example `feat: ...` or
>    `fix(client): ...`.

**Issue number**: #1495

### Change description:

`setup_auto_guards` chose the wrapper for each `GUARDED_METHODS` entry with
`is_async = "Async" in class_method`. Every provider in that table matches that naming except
Mistral, which spells its coroutine method `Chat.complete_async`, so it got
`_make_sync_guard_wrapper`. That wrapper passes the un-awaited coroutine to `_apply_postflight`,
the extractor finds no `choices`/`content`/`text` on it and returns `""`, and the pipeline
short-circuits. The coroutine is handed back unchanged, so the call still works and preflight
still runs, but the response guards never do: `PII(action="redact")` redacts
`client.chat.complete()` output and leaves `client.chat.complete_async()` output alone, and
`action="deny"` cannot block an async Mistral response.

The fix resolves the target attribute and asks it whether it is a coroutine function instead of
matching on its name. `inspect.unwrap` is needed because guards are installed after the
instrumentors, so the attribute is already a wrapt proxy at that point, and
`inspect.iscoroutinefunction` on the proxy reports `False` for `openai`'s and `anthropic`'s async
methods. Resolution errors stay inside the existing `try`, so a provider that is not installed is
still skipped with the same debug log.

Reproduction from the issue, no API key (Mistral's HTTP client swapped for an
`httpx.MockTransport`), on `main`:

```
sync : Reach me at [REDACTED:email]
async: Reach me at victim@example.com
```

and on this branch:

```
sync : Reach me at [REDACTED:email]
async: Reach me at [REDACTED:email]
```

Blast radius: the wrapper choice changes for exactly one entry. Comparing the old name check with
the new resolution for all 16 entries, with `openai`, `anthropic`, `groq`, `cohere` and
`mistralai` 1.12.4 installed, `mistralai.chat.Chat.complete_async` flips from sync to async and
every other entry is unchanged. `setup_auto_guards` logs `wrapped 14/16 provider methods` before
and after, with the same two skips.

Tests, run with `mistralai` 1.12.4 on Python 3.12:

```
$ pytest tests/test_guard_integration.py -q
...............                                                          [100%]
15 passed in 0.16s
```

With `src/openlit/guard/_integration.py` reverted to `main` and the new tests kept, the
regression test fails on the bug itself:

```
>       assert "[REDACTED:email]" in async_response.choices[0].message.content
E       AssertionError: assert '[REDACTED:email]' in 'Reach me at victim@example.com'
1 failed, 13 passed in 0.11s
```

The test uses a small in-memory provider module named the way Mistral names its methods, so it
does not need `mistralai` (which is unpinned in `tests/requirements.txt` and now resolves to 2.x,
where `mistralai.chat` no longer exists).

Also ran the hermetic suites around this area (`test_guard_pii`, `test_guard_pipeline`,
`test_guard_prompt_injection`, `test_guard_integration`, `test_genai_helpers`,
`test_instrumentor_names`, `test_context_detach`, `test_cache_pricing`): 132 passed, 1 skipped.
`pylint` reports both changed files clean at 10.00/10; the 10 pre-existing messages elsewhere in
`sdk/python` are identical before and after this change. `poetry check` and `poetry build` pass.

### Checklist

* [ ] Changes are documented

### Acknowledgment

## Summary by Sourcery

Correct automatic guard wrapper selection so asynchronous provider methods receive postflight processing regardless of their naming convention.

Bug Fixes:
- Ensure postflight guards run for Mistral-style asynchronous chat methods, including response redaction and denial behavior.

Enhancements:
- Detect provider methods as coroutine functions from their resolved attributes rather than relying on method naming conventions.

Tests:
- Add integration coverage for asynchronous methods whose names use a trailing `_async` suffix.


amanagarwal042:
Hi @aniketwaghh Thanks for the PR. I have a proposal for you and your team, Could you reach out to me at contact@openlit.io ?

PR #1497, opened 2026-08-30
fix(anthropic): guard non-streaming token counts against None

Aftabbs:
## Problem

The streaming path in `process_streaming_chat_response` already applies `or 0` guards when extracting token counts from usage dicts:

```python
# streaming – correct
scope._input_tokens = message_usage.get("input_tokens", 0) or 0
scope._output_tokens = usage.get("output_tokens", 0) or 0
scope._cache_creation_input_tokens = (
    message_usage.get("cache_creation_input_tokens", 0) or 0
)
scope._cache_read_input_tokens = usage.get("cache_read_input_tokens", 0) or 0
```

But `process_chat_response` (non-streaming) does not:

```python
# non-streaming – missing guard
scope._input_tokens = usage.get("input_tokens", 0)       # ← no or 0
scope._output_tokens = usage.get("output_tokens", 0)
scope._cache_creation_input_tokens = usage.get("cache_creation_input_tokens", 0)
scope._cache_read_input_tokens = usage.get("cache_read_input_tokens", 0)
```

`dict.get("key", default)` only returns the default when the key is **absent**. When the Anthropic API returns `{"input_tokens": null}` (observed in extended-thinking and some cached-response edge cases), `.get("input_tokens", 0)` returns `None`. That `None` then:

1. Propagates into `span.set_attribute(GEN_AI_USAGE_INPUT_TOKENS, None)` → `ValueError` from OpenTelemetry
2. Propagates into cache-token metric recording (lines 700-708) with no None guard → same crash

## Fix

Add the same `or 0` guard to all four token assignments in `process_chat_response`:

```python
scope._input_tokens = usage.get("input_tokens", 0) or 0
scope._output_tokens = usage.get("output_tokens", 0) or 0
scope._cache_creation_input_tokens = usage.get("cache_creation_input_tokens", 0) or 0
scope._cache_read_input_tokens = usage.get("cache_read_input_tokens", 0) or 0
```

This makes the non-streaming path consistent with the streaming path, which already handles this correctly.

## Testing

The fix is a one-liner guard per field; existing tests continue to pass. The guard is a no-op when the API returns an integer (normal case) and returns `0` instead of `None` when the API returns `null`.

## Summary by Sourcery

Bug Fixes:
- Prevent non-streaming Anthropic token usage processing from propagating null values into telemetry attributes and metrics.


amanagarwal042:
@Aftabbs great contribution! Btw can you reach out to me over contact@openlit.io? I have a proposal for you and your team

PR #1516, opened 2026-09-04
fix(groq): end streaming span on early break/close

### LINKED ISSUE #1514: [BUG] Groq streaming span leaks (never exported) when the caller breaks out of the stream early
**Describe the bug**

The Groq streaming instrumentation ends its span only inside the `except StopIteration`/`StopAsyncIteration` handler in `__next__`/`__anext__`. `TracedSyncStream.__exit__` / `TracedAsyncStream.__aexit__` merely forward to the wrapped stream, and there is no `close()`/`aclose()`. So when a caller breaks out of `with client.chat.completions.create(stream=True) as stream:` before the stream is exhausted — or calls `stream.close()` early, the normal pattern in agentic tool-calling — `StopIteration` never fires, the span stays recording forever and is never exported. The entire call is lost: no span, no cost, no token usage.

**Where** (`sdk/python`): `src/openlit/instrumentation/groq/groq.py` (`TracedSyncStream`, `__exit__`) and `src/openlit/instrumentation/groq/async_groq.py` (`TracedAsyncStream`, `__aexit__`).

**To reproduce**
```python
with client.chat.completions.create(model="llama-3.1-8b-instant",
                                    messages=[...], stream=True) as stream:
    for chunk in stream:
        break  # early exit -> span never ends, never exported
```

**Expected:** exactly one complete span (with token/cost attributes) is exported regardless of how the stream is left — full consumption, early `break`, or early `close()`.

**Precedent:** identical to the Anthropic instance fixed in #1461 (the reference implementation: `_streaming_response_processed` flag + `_finalize_streaming_span()` from `__exit__`/`close()`) and the OpenAI instance in #1454/#1455. Same copy-pasted idiom.

pacocartones:
Fixes #1514.

### Problem

`TracedSyncStream`/`TracedAsyncStream` in the Groq instrumentation end the span only inside the `except StopIteration`/`StopAsyncIteration` handler in `__next__`/`__anext__`. `__exit__`/`__aexit__` merely forward to the wrapped stream, and there is no `close()`/`aclose()`. So a caller that breaks out of `with client.chat.completions.create(stream=True) as stream:` early — or calls `stream.close()` (the canonical agentic tool-calling pattern) — never triggers `StopIteration`, the span never ends, and the whole call is lost (no span, no cost, no tokens).

### Fix (mirrors the merged Anthropic fix #1461)

- add `self._streaming_response_processed = False`;
- extract the `StopIteration` finalize body into a flag-guarded `_finalize_streaming_span()`;
- call it from `__next__`/`__anext__`, from `__exit__`/`__aexit__` in a `finally` when `not exc_type`, and from a new `close()`/`aclose()`;
- use `safe_detach(token)` for the OTel context token, matching the reference idiom.

### Test

New hermetic `tests/test_groq_stream_span_lifecycle.py` (7 cases) drives the real `chat`/`async_chat` factory with a synthetic groq-shaped context-manager stream — no network, no vendor SDK:
- **RED** (pre-fix): 4 failed / 3 passed — the four early-exit cases (sync/async early `break` + `close`/`aclose`) export **0 spans** (`assert 0 == 1`).
- **GREEN** (post-fix): 7 passed.

`pytest` exit 0 (7 passed); `pylint` 10.00/10 on the two src files + the test (tests follow the repo's `test_groq.py` explicit-marker convention so the pylint CI gate stays green).

## Summary by Sourcery

Finalize Groq streaming spans across all normal and early-exit paths so streamed calls consistently export complete telemetry.

Bug Fixes:
- Ensure Groq synchronous and asynchronous streaming spans are finalized when streams are closed, exited early, or exhausted, preserving usage and cost telemetry.
- Prevent streaming spans from remaining attached to the OpenTelemetry context after request setup failures or normal completion.

Enhancements:
- Make streaming span finalization idempotent across iteration, context-manager exit, and explicit close operations.
- Add synchronous and asynchronous close support to the traced Groq stream wrappers.

Tests:
- Add hermetic lifecycle coverage for Groq streaming completion, early exits, explicit close operations, context detachment, and error handling.


FBISiri (commented):
Checked this against the Anthropic reference (#1461) — the flag-guarded `_finalize_streaming_span()` plus `close()`/`aclose()` shape is the right port, and the 7-case hermetic test is the coverage this bug class needs. Three things from reading the diff:

**1. The exception exit path still leaks the span.** In `groq.py::__exit__` (and `async_groq.py::__aexit__`) finalization is guarded by `if not exc_type:`. Groq has no equivalent of Anthropic's `TracedMessageStreamManager.__exit__`, which handles the other branch explicitly — `handle_exception(self._span, exc_val)` then `span.end()` when `exc_type` is truthy. So `with client.chat.completions.create(stream=True) as s: raise RuntimeError(...)`, or any API error propagating out of `__next__`, still leaves the span recording and never exported — the same symptom as #1514, just on a different path. Suggest mirroring that else-branch here and adding an 8th case asserting one exported span with error status.

**2. Streamed tool calls produce spans with no tool attributes.** The PR motivates `close()` as the canonical agentic tool-calling pattern, but `process_chunk()` in `groq/utils.py` only reads `delta.content` and `delta.reasoning`; `scope._tools` stays `None` on the streaming path, so the tool block in `common_chat_logic` (hardened by merged #1452) never fires for streams. No conflict with #1452 — different file, and arguably out of scope here — but worth a follow-up issue, since it's exactly the scenario this PR rescues.

**3. Overlap with #1517.** The mistral wrappers there are line-for-line the same logic, and both PRs carry a byte-identical hunk against `tests/test_anthropic_stream_span_lifecycle.py`. Git will likely merge that cleanly, but it's duplicated churn in two PRs and turns into a conflict if either gets rebased/edited — cleaner to drop it from one. Longer term, with anthropic/groq/mistral all needing it, `_finalize_streaming_span` + the processed flag look extractable into `__helpers`.

amanagarwal042:
Hey @pacocartones PR looks great!

QQ I have a proposal for you and your team, Could you reach out to me over contact@openlit.io?

PR #1517, opened 2026-09-04
fix(mistral): end streaming span on early break/close

### LINKED ISSUE #1515: [BUG] Mistral streaming span leaks (never exported) when the caller breaks out of the stream early
**Describe the bug**

The Mistral streaming instrumentation ends its span only inside the `except StopIteration`/`StopAsyncIteration` handler. `TracedSyncStream.__exit__` / `TracedAsyncStream.__aexit__` only forward to the wrapped stream, and there is no `close()`/`aclose()`. A caller that breaks out of the `with` block before exhaustion, or closes the stream early, never triggers `StopIteration`, so the span stays recording and is never exported — the whole call (span, cost, tokens) is lost.

**Where** (`sdk/python`): `src/openlit/instrumentation/mistral/mistral.py` (`stream()` -> `TracedSyncStream`, `__exit__`) and `src/openlit/instrumentation/mistral/async_mistral.py` (`async_stream()` -> `TracedAsyncStream`, `__aexit__`). This is the early-close leak, **distinct from** the async postflight bug in #1495/#1496.

**To reproduce**
```python
with client.chat.stream(model="mistral-small-latest", messages=[...]) as stream:
    for event in stream:
        break  # early exit -> span never ends, never exported
```

**Expected:** exactly one complete span (with token/cost attributes) exported on every exit path — full consumption, early `break`, or early `close()`.

**Precedent:** same vein as the merged Anthropic fix #1461 and OpenAI #1454/#1455.

pacocartones:
Fixes #1515.

### Problem

`TracedSyncStream`/`TracedAsyncStream` in the Mistral instrumentation end the span only inside the `except StopIteration`/`StopAsyncIteration` handler. `__exit__`/`__aexit__` only forward to the wrapped stream, and there is no `close()`/`aclose()`. A caller that breaks out of `with client.chat.stream(...) as stream:` early — or closes it early — never triggers `StopIteration`, so the span never ends and the whole call is lost (no span, no cost, no tokens). This is the early-close leak, **distinct from** the async postflight bug in #1495/#1496.

### Fix (mirrors the merged Anthropic fix #1461)

- add `self._streaming_response_processed = False`;
- extract the `StopIteration` finalize body into a flag-guarded `_finalize_streaming_span()`;
- call it from `__next__`/`__anext__`, from `__exit__`/`__aexit__` in a `finally` when `not exc_type`, and from a new `close()`/`aclose()`;
- use `safe_detach(token)` for the OTel context token, matching the reference idiom.

### Test

New hermetic `tests/test_mistral_stream_span_lifecycle.py` (7 cases) drives the real `stream`/`async_stream` factory with a synthetic mistral-shaped context-manager stream — no network, no vendor SDK:
- **RED** (pre-fix): 4 failed / 3 passed — the four early-exit cases export **0 spans**.
- **GREEN** (post-fix): 7 passed.

`pytest` exit 0 (7 passed); `pylint` 10.00/10. Only the streaming early-close leak is touched; the async postflight area (#1495/#1496) is untouched.

## Summary by Sourcery

Finalize Mistral streaming telemetry reliably across normal completion and early termination paths.

Bug Fixes:
- Ensure Mistral synchronous and asynchronous streaming spans are finalized when streams are exited or closed before exhaustion, preventing lost spans and usage data.
- Prevent duplicate streaming span finalization across exhaustion, early exit, and explicit close operations.

Enhancements:
- Use safe context detachment for Mistral streaming instrumentation.

Tests:
- Add hermetic regression coverage for synchronous and asynchronous early exits, explicit close operations, full consumption, token usage, and context cleanup.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/mistral/async_mistral.py
+++ b/sdk/python/src/openlit/instrumentation/mistral/async_mistral.py
@@ -10,6 +10,7 @@ from openlit.__helpers import (
     set_server_address_and_port,
     record_completion_metrics,
     record_embedding_metrics,
+    safe_detach,
 )
 from openlit.instrumentation.mistral.utils import (
     process_chunk,
@@ -149,13 +150,25 @@ def async_stream(
             self._tbt = 0
             self._server_address = server_address
             self._server_port = server_port
+            self._streaming_response_processed = False
 
         async def __aenter__(self):
             await self.__wrapped__.__aenter__()
             return self
 
         async def __aexit__(self, exc_type, exc_value, traceback):
-            await self.__wrapped__.__aexit__(exc_type, exc_value, traceback)
+            try:
+                await self.__wrapped__.__aexit__(exc_type, exc_value, traceback)
+            finally:
+                if exc_type:
+                    self._streaming_response_processed = True
+                    handle_exception(self._span, exc_value)
+                    if self._span.is_recording():
+                        self._span.end()
+                else:
+                    # A break before exhaustion never hits StopAsyncIteration,
+                    # so the span would leak without normal-exit finalization.
+                    self._finalize_streaming_span()
 
         def __aiter__(self):
             return self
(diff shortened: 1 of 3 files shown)


pacocartones:
The token-accounting review finding is addressed in `c1272036`: early-close paths now assert exact `0/0` usage and fully consumed streams assert `10/5`. The focused tests and the full PR CI are green, and Sourcery approved the updated head. Ready for maintainer review.

PR #1520, opened 2026-09-04
fix(onboarding): correct empty step 3 database config UI

### LINKED ISSUE #1518: Bug: Onboarding in inescapable loop on :latest.. Step 3 renders no db config ui option.
### Component

OpenLIT

### What happened?

On a fresh Docker Compose deployment using ghcr.io/openlit/openlit:latest, the onboarding wizard skips step 2 and lands on step 3 ("connect a database") with no interactive elements at all — no "Add new config" button, no modal, no form. Every other route (/home, /dashboard, /requests) redirects back to /onboarding, so the instance is unusable.

Pinning to a tagged release, with an otherwise identical compose file, ClickHouse volume, and configuration, renders the same step correctly with an "Add new config" control available. This appears to be a regression in latest. Works great and as intended. 

The backend is healthy throughout: ClickHouse contains the complete application schema, credentials authenticate, and OTLP ingestion works — traces from a LiteLLM proxy land in otel_traces the whole time the UI is inaccessible.

### Steps to reproduce

Openlit latest
Clickhouse 24.4.1
Host: Docker desktop, Windows 11, WSL2 Backend
Deploy: Docker compose from repo's docker-compose.yml
telemetry source: LiteLLM Latest:otel callback over http (local net)

Deviation from Norms:

Existing openwebui instance is pinned to tcp:3000 so openlit compose WAS changes to 3001 to avoid port conflict.

Wait for INIT_DB_* to seed ClickHouse connection. 

Login, create account. Add an arbitrary org name 'My cool org'. UI Skips step 2, bypass step 3 DB onboard. Fails in an infinite loop. 

Wiping the openlit-data volume and re-running does not help — onboarding restarts and stalls at the same point.

[Onboarding-Loop-Logs.txt](https://github.com/user-attachments/files/31821690/Onboarding-Loop-Logs.txt)

### Expected behavior

Should log in, create org, skip project due to compose seed, then be dropped at a Step 3 onboarding ready to add your Clickhouse connection.

### Environment

Open Lit Latest  (Resolved when i back rev' to 1.24
Clickhouse 24.4.1
Docker desktop , WSL2 Ubuntu image. Flat network

### Additional context

Workaround was to just pin the compose to a previous version instead of :latest. (2.0)

yaml
image: ghcr.io/openlit/openlit:1.24.0
bash
docker compose down
docker volume rm <project>_openlit-data
docker compose pull openlit
docker compose up -d

The ClickHouse volume can be retained; only application state resets. Onboarding then renders correctly and completes. With telemetry intact. 

### Pre-submission checklist

None

BetterAndBetterII:
## Summary
- Onboarding step 3 no longer links to `/organisation/project/:id/connectors` (which redirects incomplete users back to `/onboarding`), so the "Add database config" control is empty and the wizard loops.
- Embed `DatabaseConfigPage` inline on step 3 and whitelist `/api/db-config`, `/api/clickhouse`, and `/api/project/environment` while onboarding is incomplete so the ClickHouse form can load and save.

## Test plan
- [x] `npx jest` for `onboarding-page`, `route`, and `check-auth` tests (54 passed)
- [ ] Fresh compose with `:latest`: create org → step 3 shows Add database config → form opens → save ClickHouse → finish setup

Fixes #1518

## Summary by Sourcery

Enable database configuration directly within onboarding so setup can be completed without redirect loops.

Bug Fixes:
- Fix onboarding step 3 so users can add and save a database configuration without being redirected out of the setup flow.

Enhancements:
- Embed the database configuration experience directly in onboarding and determine setup completion from available database configurations.
- Allow required database configuration, ClickHouse, and project environment APIs during incomplete onboarding.

Build:
- Make Docker dependency installation resilient to expired repository metadata timestamps in CI.

Tests:
- Add coverage for the inline onboarding database configuration flow and its pre-completion API access.


AmanAgarwal041 (dismissed):

AmanAgarwal041:
Thank you for this — the step 3 database-config form and onboarding API allowlist are the right fix.

We re-ran CI on this PR, then landed the change on current `main` in https://github.com/openlit/openlit/pull/1534 with one extra follow-up: home and chat were still checking `/api/connectors`, so finishing setup could bounce users back to `/onboarding`. Your commit is preserved there.

AmanAgarwal041:
Pushed a follow-up to this branch for the **Build / Container** failure: `debian-security` `bullseye-security` InRelease was expired during `apt-get update` in `src/Dockerfile`.

The image now runs `apt-get update -o Acquire::Check-Valid-Until=false` so docker compose can still install runtime packages when that timestamp is briefly stale.

PR #1522, opened 2026-09-05
fix: fail fast when clickhouse migrations fail

### LINKED ISSUE #1519: Bug: ClickHouse migrations report success after partially failing on an auth error, leaving the application schema incomplete
### Component

OpenLIT

### What happened?

When OpenLIT cannot authenticate to ClickHouse partway through initialisation, it creates the otel_* and openlit_controller_* tables, fails to create any application tables, and then reports success and continues — logging ClickHouse tables created, running the data migration, and reporting Seeding Dashboards Completed (seeded 4, skipped 0).

The failure only surfaces later as UNKNOWN_TABLE errors at runtime, in a component unrelated to the actual cause. This makes the root cause very difficult to identify.

### Steps to reproduce

A ClickHouse container was first initialised without a .env file present, or with a path fail or other defect in .env.

So it took the compose fallback password. A .env with a different password was added afterwards. **_Because CLICKHOUSE_PASSWORD is only honoured on first initialisation of the data directory_**, OpenLIT was then configured with credentials ClickHouse did not have.

This is user error and I'm not reporting it as a bug. The bug is that OpenLIT's response to it was to report success making triage of the subsequent failure difficult.

SHOW TABLES FROM openlit at this point returned only the six openlit_controller_* tables and the nine otel_* tables — no application tables at all.

This led me to go back and repull the previously successful logs, and renumerate the chain of events, where I discovered the mismatch env. 

[DB-Conn-Fail.txt](https://github.com/user-attachments/files/31821869/DB-Conn-Fail.txt)

### Expected behavior

Fail loudly on authentication errors during ClickHouse migration rather than continuing.
Don't log ClickHouse tables created unless they were.
Verify expected tables exist before seeding dashboards, and report which are missing.
Surface the underlying auth error at the point of failure, rather than letting it emerge later as a UNKNOWN_TABLE error in an unrelated feature.

A simple  "could not authenticate to ClickHouse — migrations aborted" message would have saved a considerable amount of debugging time.

### Environment

Openlit 1.24

### Additional context

_No response_

### Pre-submission checklist

None

mikemikimike:
## Summary

ClickHouse startup migrations could log success after one or more statements failed. On a partially initialized database, the dashboard seed then ran against missing tables and left the application schema incomplete.

## Changes

- Validate every migration result before allowing the next dependency group to run.
- Preserve the existing legacy success result shapes while surfacing reported errors and incomplete query batches.
- Skip built-in dashboard seeding when dashboard table creation did not complete.
- Verify all expected dashboard tables before seeding and report any missing table names.
- Add regression coverage for failed migration orchestration, missing-table reporting, and dashboard seed suppression.

## Tests

- `npm test -- --runInBand src/__tests__/clickhouse/migrations/migrations-index.test.ts src/__tests__/clickhouse/migrations/create-custom-dashboards-migration.test.ts`
- `npm run lint`
- `npm run build`
- Full Jest coverage: 3,209 passed; 4 tests in 14 suites retain unrelated Windows/JSDOM baseline failures in Prisma initialization, path, and locale assertions.

## Compatibility / Known limitations

Existing migration records remain idempotent, and failed migrations remain unrecorded so a later startup can retry them. ClickHouse does not provide a cross-table transaction for the independent startup group, so already-created independent tables are intentionally retained and safely retried.

## Issue

Closes #1519

## Summary by Sourcery

Make ClickHouse startup migrations fail fast and prevent dashboard seeding against incomplete schemas.

Bug Fixes:
- Fail ClickHouse startup migration orchestration immediately when a migration reports errors, incomplete query execution, or no successful result.
- Prevent built-in dashboard seeding when dashboard table creation fails or expected tables are missing.

Enhancements:
- Preserve supported legacy migration result formats while validating migration completion and reporting actionable failures.
- Verify all required dashboard tables before allowing dashboard seeding to proceed.

Tests:
- Add regression coverage for migration failure handling, missing dashboard table reporting, and seed suppression.


amanagarwal042:
@mikemikimike This was great! Thanks for this
QQ I have a proposal for you and your team that might be useful, Can you reach out to me at contact@openlit.io?

PR #1524, opened 2026-09-05
docs: add The Grid integration page

CastilloLuis:
## What

Adds a docs page for **The Grid** to the SDK integrations list, following the Featherless / Krutrim / NVIDIA NIM template for OpenAI-SDK-compatible providers.

Three registration points, no SDK changes:
- `docs/latest/sdk/integrations/thegrid.mdx` (new)
- one nav string in `docs/docs.json`
- a `<Card>` in `docs/latest/sdk/integrations/overview.mdx`

## Why no code

The Grid is an OpenAI-compatible inference marketplace, so `openlit.init()` already traces it through the existing OpenAI instrumentor — `SERVER_ADDRESS` comes out as `api.thegrid.ai` and token usage is recorded normally. There's nothing to add under `sdk/python/src/openlit/instrumentation/`.

## Verified

Checked against the live API with `openai-python 2.48.0`: sync and streaming chat completions both work, and `usage` is populated (prompt/completion tokens) so cost and token metrics have real input.

## On pricing

I deliberately did **not** add entries to `assets/pricing.json`. The Grid's instruments are market-priced and move intraday, so a static per-token float would be wrong shortly after merge. The right home for this is the `pricing_json` URL hook in `openlit.init()` — we'll host an OpenLIT-shaped pricing endpoint on our side rather than ask you to carry numbers that go stale.

Disclosure: I work on The Grid. Happy to trim the framing if it reads too vendor-ish.

## Summary by Sourcery

Add The Grid to the SDK integrations documentation and explain how to monitor it with OpenLIT.

New Features:
- Add documentation for monitoring The Grid through its OpenAI-compatible SDK integration, including setup guidance and pricing configuration details.

Enhancements:
- Register The Grid in the SDK integrations navigation and overview page.

Documentation:
- Document The Grid integration capabilities, supported OpenAI SDK usage, token and latency monitoring, and custom pricing configuration.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -318,6 +318,7 @@
                   "latest/sdk/integrations/digitalocean-gradient",
                   "latest/sdk/integrations/assemblyai",
                   "latest/sdk/integrations/featherless",
+                  "latest/sdk/integrations/thegrid",
                   "latest/sdk/integrations/reka",
                   "latest/sdk/integrations/krutrim",
                   "latest/sdk/integrations/titan-ml",
--- a/docs/latest/sdk/integrations/overview.mdx
+++ b/docs/latest/sdk/integrations/overview.mdx
@@ -734,6 +734,32 @@ Start integrating and monitoring your applications with OpenLIT in just one line
 
   </Card>
   <Card
+      title="The Grid"
+      href="/latest/sdk/integrations/thegrid"
+      icon={
+        <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+          <path d="M4 3L7.76471 6.40541V8.35135H4V3Z" fill="currentColor"/>
+          <path d="M4 21L7.76471 17.5946V15.6486H4V21Z" fill="currentColor"/>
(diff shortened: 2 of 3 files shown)


CastilloLuis:
Thanks, both findings were correct and I've pushed fixes.

**OpenAI SDK version.** Corrected to `>=1.92.0` to match the `openai = ">=1.92.0,<3.0.0"` constraint in `sdk/python/pyproject.toml`.

Worth flagging separately: I took `>=1.13.0` from the existing template, and the same stale claim appears on the sibling pages, at least `featherless.mdx:14` and `krutrim.mdx:14`. I left those alone to keep this PR scoped to The Grid, but they'd need the same correction. Happy to fix them here or in a follow-up, whichever you prefer.

**Cost tracking.** Also correct. The Grid is market-priced and publishes no per-token rate, so there is no entry in the bundled `pricing.json` and no cost can be computed. I removed the cost claim from the page description and added a warning that states this plainly, pointing at the `pricing_json` hook in `openlit.init()` for users who want to supply their own rates.

I deliberately did not add The Grid to `assets/pricing.json`: any static per-token figure would be wrong shortly after merge.

Issue #1537, opened 2026-09-08
Reasoning-token subset: apply “never sum the subset into its total” to every provider path (checklist + fixture)


roy-tong:
## Summary

PR #1476 merged the base invariant for the OpenAI/Responses paths: reasoning tokens are inside output_tokens, so they must never be summed on top. The maintainer note on that PR flagged the same double-count still exists on other providers' GEN_AI_CLIENT_TOKEN_USAGE paths. This issue is the checklist to close that class, with one fixture to pin it.

## Invariant (AgentMeasure reasoning-subset rule)

For each provider path:
- output_tokens reported = the provider total (already includes reasoning);
- - reasoning tokens, if surfaced, are a separate attribute that is a facet of output — never an addend;
- - - any "completed / non-reasoning" figure is derived as output − reasoning and labeled derived.
## Checklist

- [ ] Audit each SDK/package path building usage from provider responses: is reasoning emitted as a separate attribute only (no addition)?
- [ ] - [ ] Where dashboards compute "completed" or "non-reasoning" tokens: is it a derivation (output − reasoning), never a sum?
- [ ] - [ ] Fixture: synthetic provider usage (output 1,000 / reasoning 700) → assert recorded output = 1,000 and that no downstream aggregate can reach 1,700; missing-reasoning case still parses (UNPROVABLE, not a guess).
Happy to attach the fixture (we already keep this shape in AgentMeasure's corpus) and review the paths one by one. Thanks again for the earlier merge and for asking what the solution should be.

roy-tong:
Fixture for the invariant, inline (we keep this shape in AgentMeasure's corpus):

```json
{
  "provider": "openai",
  "usage": {
    "output_tokens": 1000,
    "output_tokens_details": { "reasoning_tokens": 700 }
  },
  "expected": {
    "output_recorded": 1000,
    "reasoning_emitted_as_facet": 700,
    "forbidden_downstream_aggregate": 1700,
    "missing_reasoning_case": "parses with reasoning absent - report UNPROVABLE, never 0"
  }
}
```

Assertions per path: output == provider total; reasoning never added to output; a derived "completed tokens" figure equals output - reasoning and is labeled derived; a provider that stops sending reasoning_details must surface as missing (UNPROVABLE), not as 0. Happy to drop this as a real file under a fixtures dir on whichever path you want to pin first.

roy-tong:
[...]

PR #1476, opened 2026-08-19
feat: record reasoning output tokens without double counting token usage

roy-tong:
Closes #1474

**Problem:** PR #1018 correctly aligned gen_ai.client.token.usage with the OTel GenAI semantic convention by limiting token_type to input/output. Provider-reported reasoning tokens are currently not exposed as a normalized attribute.

**Change (scope: Python/OpenAI first):**
- Extract completion_tokens_details.reasoning_tokens in chat (stream + non-stream) and responses paths
- - Expose as gen_ai.usage.reasoning.output_tokens (OTel GenAI semconv, experimental; SHOULD be included in gen_ai.usage.output_tokens)
- - gen_ai.client.token.usage stays input/output only, no third token type
**Invariant test:** output_tokens = 1000, reasoning.output_tokens = 700, client.token.usage{output} = 1000 (no 1700 double count).

**Tests:** 4 new + 108 regression passed. pylint 10.00/10.

## Summary by Sourcery

Expose OpenAI reasoning output tokens through the OTel GenAI semantic convention without inflating output token usage totals.

New Features:
- Record normalized OpenTelemetry reasoning output token usage for OpenAI chat completions and Responses API calls, including streaming responses.

Bug Fixes:
- Prevent reasoning tokens from being double-counted in output token usage metrics while retaining the existing output total.

Enhancements:
- Add defensive handling for missing or malformed token detail objects and preserve the legacy reasoning token attribute for compatibility.

Tests:
- Add coverage for chat, streaming, and Responses API reasoning token extraction, zero or absent reasoning tokens, malformed usage details, and the no-double-counting invariant.

PR #1543, opened 2026-09-09
fix(openai): measured-zero reasoning vs unknown + derived completed tokens (#1537 chat-completions path)

roy-tong:
## What

First checklist item of #1537, scoped to the OpenAI chat-completions path (chat + streaming chat; the shared emission helper also keeps the responses path consistent).

## Why

After #1476, reasoning is correctly emitted as a subset facet. Two gaps remained on this path:

1. Measured zero collapsed into unknown. extract_reasoning_tokens fell back to 0 when completion_tokens_details was missing, and the emission guard was greater-than 0, so an explicit reasoning_tokens: 0 was omitted too. A consumer could not distinguish no-reasoning from unknown (the #1537 fixture UNPROVABLE case).
2. No derived completed figure. Dashboards wanting completed / non-reasoning tokens had to compute it themselves, with no label marking it derived.

## Changes

- extract_reasoning_tokens returns None when the details object is absent (unknown), keeping an explicit 0 as a measurement.
- New marker gen_ai.usage.reasoning.output_tokens.reported (true/false) makes the unknown state explicit instead of guessing 0.
- An explicit 0 now emits the facet attribute with value 0 (behavior change, tests updated).
- New gen_ai.usage.derived.completed_output_tokens = output minus reasoning, namespaced under derived.* so it can never be mistaken for a provider-reported number.
- Legacy alias gen_ai.usage.reasoning_tokens stays nonzero-only: pre-OTel consumers see no change.
- Regression fixture from #1537 added as tests/test_openai_reasoning_subset_fixture.py: the 1000/700 subset case, the forbidden 1700 aggregate check, missing-details unknown case, measured-zero case, and the streaming path.

## Test plan

- pytest tests/test_openai_reasoning_tokens.py tests/test_openai_reasoning_subset_fixture.py = 11 passed
- Full collectible suite locally: 630 passed; the 5 failures in test_anthropic_stream_span_lifecycle / test_google_adk / test_litellm_reproduction reproduce on a clean main checkout in this environment (optional-dependency version drift), unrelated to this change.

Fixes the chat-completions items of #1537.

## Summary by Sourcery

Distinguish reported reasoning-token measurements from unknown usage while exposing derived completed output tokens without changing total output accounting.

New Features:
- Add derived completed output-token usage and an explicit marker indicating whether reasoning-token usage was reported.

Bug Fixes:
- Preserve the distinction between measured zero reasoning tokens and unknown reasoning usage across OpenAI chat and streaming paths.
- Prevent reasoning subsets from being double-counted with total output tokens.

Enhancements:
- Keep the legacy reasoning-token alias backward compatible by emitting it only for nonzero values.
- Apply consistent reasoning subset emission behavior to the OpenAI responses path.

Tests:
- Add regression coverage for measured zero, unknown usage, subset accounting, derived completed tokens, and chat/responses streaming scenarios.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/openai/utils.py
+++ b/sdk/python/src/openlit/instrumentation/openai/utils.py
@@ -105,16 +105,67 @@ def extract_reasoning_content(payload):
 
 
 def extract_reasoning_tokens(usage, details_key):
-    """Return reasoning tokens (a subset of output tokens) from a usage dict.
+    """Return reasoning tokens (a subset of output tokens) from a usage dict,
+    or None when the provider did not report the details object.
 
     Works for both chat completions (``completion_tokens_details``) and the
-    responses API (``output_tokens_details``) and falls back to 0 whenever the
-    details object is missing or not a dict.
+    responses API (``output_tokens_details``). A missing or non-dict details
+    object means "unknown" and stays distinguishable from an explicit
+    ``reasoning_tokens: 0`` measurement (None vs 0): a reported 0 is a
+    measurement, an absent object is an unknown, and the two must not collapse
+    into the same value.
     """
-    details = (usage or {}).get(details_key) or {}
+    details = (usage or {}).get(details_key)
     if not isinstance(details, dict):
-        details = {}
-    return details.get("reasoning_tokens", 0) or 0
+        return None
+    value = details.get("reasoning_tokens")
+    if not isinstance(value, (int, float)) or isinstance(value, bool):
+        return None
+    return value
+
+
+def set_reasoning_subset_attributes(span, output_tokens, reasoning_tokens):
+    """Emit reasoning-token attributes under the subset invariant.
+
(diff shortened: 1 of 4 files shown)


roy-tong:
Closing the loop on the merge — thank you. The measured-zero-vs-unknown distinction this lands (reasoning tokens reported as an actual 0 stay 0, absent stays unknown rather than defaulting) is exactly the absent≠zero discipline; the same fixture family with both shapes side by side is in the pack below for regression.
---

<sub>Repro fixtures: [AgentMeasure conformance pack](https://github.com/roy-tong/AgentMeasure/tree/main/conformance/pack) — plain JSONL + expected totals, no runtime install. Found during an [open ecosystem audit](https://github.com/roy-tong/AgentMeasure/blob/main/campaigns/measurement-casebook.md#ecosystem-audit-33-verified-usage-accounting-findings-2026-09-0607) of usage-accounting tools.</sub>

Issue #1454, opened 2026-08-13
Bug: OpenAI streaming spans are never ended if the stream is closed early


AmirF194:
### Component

OpenLIT Python SDK

### What happened?

`TracedSyncStream`/`TracedAsyncStream` in the OpenAI instrumentation (chat completions and the Responses API, sync and async) only end the OTel span they created inside the `StopIteration`/`StopAsyncIteration` handler in `__next__`/`__anext__`. `__exit__`/`__aexit__` just forward to the wrapped OpenAI stream's own exit, and an explicit `.close()` call bypasses the wrapper entirely via `__getattr__`, since `TracedSyncStream`/`TracedAsyncStream` never define their own `close()`.

So any caller that stops consuming a stream before it raises `StopIteration` (a `with ... as stream:` block that `break`s out of the `for` loop, or a bare `stream.close()`) never triggers `span.end()`. The span is created via `tracer.start_span(...)`, not a `with`-managed context, so it is never handed to the exporter. The whole call, including any prompt/completion content and token usage that had already streamed in before the early stop, is invisible to the observability backend: no span, no cost, no record the call happened at all. This is a normal pattern for tool-calling/agentic loops that stop consuming a stream once they have what they need.

### Steps to reproduce

1. Instrument the OpenAI client with `openlit.init(...)`.
2. Call `client.chat.completions.create(stream=True, ...)` inside `with client.chat.completions.create(...) as stream:`.
3. Iterate a few chunks with `for chunk in stream:` and `break` before the stream is exhausted (or call `stream.close()` directly instead of using `with`).
4. Check the configured span exporter: no span for that call is ever exported.

### Expected behavior

The span should still be ended (with whatever partial content/usage had already been captured) whenever the stream is closed early, the same way it is when the stream is fully consumed to `StopIteration`.

### Additional context

Affects all 4 occurrences of the pattern: `sdk/python/src/openlit/instrumentation/openai/openai.py` (`TracedSyncStream` in both `chat_completions()` and `responses()`) and the async twins in `sdk/python/src/openlit/instrumentation/openai/async_openai.py`. Introduced with the original class-based stream wrapper (PR #503) and copied into the Responses API wrapper (PR #658); it has been present since then.

feiiiiii5:
[...]

PR #1546, opened 2026-09-10
fix(python-sdk): end bedrock streaming span on early break/close

### LINKED ISSUE #1545: Bug: Bedrock converse_stream spans are never ended if the stream is closed early
### Component

OpenLIT Python SDK

### What happened?

`TracedSyncStream` in the Bedrock instrumentation (`sdk/python/src/openlit/instrumentation/bedrock/bedrock.py`, `converse_stream`) only ends the span it created inside the `StopIteration` handler in `__next__`. `__exit__` just forwards to the wrapped botocore stream's own exit if it has one, and there's no `close()` of its own, so an explicit `.close()` falls through `__getattr__` to the wrapped response instead.

So any caller that stops consuming a `converse_stream` response before `StopIteration`, a `with ... as stream:` block that `break`s out of the loop, or a bare `.close()`, never ends the span. It was created with `tracer.start_span(...)`, not a `with`-managed context, so it's never exported. The whole call, tokens, cost, everything, is invisible to the observability backend. This is the normal shape of an agentic tool-calling loop that stops reading a stream once it has what it needs.

This is the same defect already fixed for Anthropic (#1461) and reported for OpenAI (#1454) and Groq/Mistral (#1514/#1515), just not yet reported for Bedrock.

### Steps to reproduce

1. Instrument the client with `openlit.init(...)`.
2. Call `client.converse_stream(...)` inside `with client.converse_stream(...) as stream:`.
3. Iterate a few chunks with `for chunk in stream:` and `break` before the stream is exhausted.
4. Check the configured span exporter, no span for that call is ever exported.

### Expected behavior

The span should still be ended (with whatever partial content/usage had already been captured) when the stream is closed early, the same way it is once the stream is fully consumed to `StopIteration`.

### Environment

- Current main, commit `6b3146685a3f9483db8bc0822f5d7c33a54a282f`
- Python 3.12

### Additional context

I confirmed this with a real OpenTelemetry SDK `InMemorySpanExporter`: driving `converse_stream` through the real `TracedSyncStream` with a fake event stream of two chunks, breaking after the first chunk exports zero spans, and letting it run to `StopIteration` exports one. Happy to send a PR, the fix looks like the same shape #1461 used for Anthropic: a `_finalize_streaming_span()` helper with an idempotency flag, called from both `__exit__`/`close()` and the existing `StopIteration` branch, so an early exit still finalizes exactly once.

wangzhengzhuo05:
**Issue number**: #1545

### Change description:

`TracedSyncStream` in the Bedrock `converse_stream` instrumentation ended its span only inside the `StopIteration` branch of `__next__`. `__exit__` merely forwarded to the wrapped botocore stream and there was no `close()`, so any caller that stopped consuming the stream before exhaustion — a `with ... as stream:` block that `break`s out of the loop, or a bare `stream.close()` — left the span recording forever and it was never exported. The whole call (tokens, cost, response) was invisible to the observability backend. This is the normal shape of an agentic tool-calling loop that stops reading a stream once it has what it needs, and it is the Bedrock instance of the same defect already fixed for Anthropic (#1461) and reported for OpenAI (#1454) / Groq-Mistral (#1514/#1515).

The fix mirrors the merged Anthropic implementation (#1461):

- Added an idempotency flag `_streaming_response_processed` (init `False`).
- Extracted the existing `StopIteration` finalization into `_finalize_streaming_span()`, guarded by the flag so the span is completed and exported exactly once.
- `__exit__` now finalizes on clean exit (when no exception escaped the `with` block), so an early `break` still exports the span.
- Added `close()` which closes the wrapped stream (when it has one) and finalizes the span in a `finally`.

### Verification

- Added `sdk/python/tests/test_bedrock_stream_span_lifecycle.py` driving the real `converse_stream` wrapper factory with a synthetic Bedrock event stream and an `InMemorySpanExporter`. Three exit paths each export exactly one finished span:
  - early `break`/leave of the `with` block,
  - full consumption to `StopIteration` (still exactly one span — idempotency),
  - explicit `stream.close()` inside the block.
- `pytest sdk/python/tests/test_bedrock_stream_span_lifecycle.py -q` → `3 passed`.
- Mutation check: reverting only the exit-time finalize makes `test_sync_early_break_inside_with_ends_span` fail with `len == 0` (span leaked); restoring it passes again.
- Existing `test_anthropic_stream_span_lifecycle.py` sync tests still pass (the 2 async cases require `pytest-asyncio`, which is absent in this environment and untouched).

### Checklist

* [ ] Changes are documented

### Acknowledgment

### AI-generated code disclosure

Implemented with AI assistance (OpenCode + muse-spark 1.3) from a self-contained task specification, then reviewed and verified locally for correctness, test coverage, and scope (only `bedrock.py` and the new test file changed; no new dependencies).

Fixes #1545

## Summary by Sourcery

Ensure Bedrock streaming telemetry is completed on every stream lifecycle exit path.

Bug Fixes:
- Ensure Bedrock streaming spans are finalized and exported when streams are exited early, explicitly closed, fully consumed, or terminated by an exception.

Enhancements:
- Make streaming span finalization idempotent so each response produces exactly one completed span.

Tests:
- Add lifecycle regression tests covering early exit, full consumption, explicit close, and exception exit paths.

### DIFF
--- a/sdk/python/src/openlit/instrumentation/bedrock/bedrock.py
+++ b/sdk/python/src/openlit/instrumentation/bedrock/bedrock.py
@@ -191,9 +191,12 @@ def converse_stream(
                     self.__wrapped_stream.__exit__(exc_type, exc_value, traceback)
             finally:
                 # Finalize on every exit: a break before exhaustion never
-                # hits StopIteration, so the span would leak otherwise.
-                if not exc_type:
-                    self._finalize_streaming_span()
+                # hits StopIteration, so the span would leak otherwise. Also
+                # finalize on exception exits (exc_type set) — the flag in
+                # _finalize_streaming_span keeps a double call a no-op, so the
+                # span is still exported even when the caller aborts the
+                # with-block with an error that never reaches StopIteration.
+                self._finalize_streaming_span()
 
         def __iter__(self):
             return self
--- a/sdk/python/src/openlit/instrumentation/bedrock/bedrock.py
+++ b/sdk/python/src/openlit/instrumentation/bedrock/bedrock.py
@@ -178,6 +178,7 @@ def converse_stream(
             self._server_address = server_address
             self._server_port = server_port
             self._event_provider = event_provider
+            self._streaming_response_processed = False
 
         def __enter__(self):
             if hasattr(self.__wrapped_stream, "__enter__"):
@@ -185,8 +186,14 @@ def converse_stream(
             return self
 
         def __exit__(self, exc_type, exc_value, traceback):
-            if hasattr(self.__wrapped_stream, "__exit__"):
-                self.__wrapped_stream.__exit__(exc_type, exc_value, traceback)
+            try:
+                if hasattr(self.__wrapped_stream, "__exit__"):
+                    self.__wrapped_stream.__exit__(exc_type, exc_value, traceback)
+            finally:
+                # Finalize on every exit: a break before exhaustion never
+                # hits StopIteration, so the span would leak otherwise.
+                if not exc_type:
+                    self._finalize_streaming_span()
 
         def __iter__(self):
             return self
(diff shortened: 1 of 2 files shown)


wangzhengzhuo05:
Closing this out.

It has been waiting 8 days without any maintainer review, and we are winding down this contribution effort — a stale PR sitting in the queue costs maintainers attention without giving them a decision to make, so it is better closed than left open indefinitely.

If the direction is useful, the branch is intact on the fork (`fix/bedrock-stream-span-finalize`, +210/−18 across 2 file(s)) and can be picked up as-is.

What it contains: *fix(python-sdk): end bedrock streaming span on early break/close* — **Issue number**: #1545

It was verified locally before opening (target test suite green, plus a mutation check that reverts only the fix and confirms the new regression test fails). Thanks for the time.

PR #1566, opened 2026-09-13
fix(together): bind instrumentation to together 2.x resource classes

### LINKED ISSUE #1565: Bug: Together AI instrumentation binds to nothing on together 2.x
### Component

OpenLIT Python SDK

### What happened?

On `together` 2.x, openlit's Together AI instrumentation binds to nothing. No spans, no metrics — the integration is completely inert.

`instrumentation/together/__init__.py` wraps four targets, and every one of them names a class that was renamed in `together` 2.0:

| openlit wraps | actual in `together` 2.33.2 |
| --- | --- |
| `together.resources.chat.completions.ChatCompletions.create` | `CompletionsResource.create` |
| `together.resources.chat.completions.AsyncChatCompletions.create` | `AsyncCompletionsResource.create` |
| `together.resources.images.Images.generate` | `ImagesResource.generate` |
| `together.resources.images.AsyncImages.generate` | `AsyncImagesResource.generate` |

The modules still import fine — only the classes moved — so nothing downstream breaks visibly.

Unlike the Mistral case in #1536, this does raise. `together.resources.chat.completions` is importable, so `wrap_function_wrapper` resolves eagerly instead of deferring to a post-import hook:

```
PathResolutionError: unable to resolve attribute 'ChatCompletions' in path
'ChatCompletions.create' on <module 'together.resources.chat.completions'>
```

But `instrument_if_available` in `openlit/__init__.py` catches it:

```python
except Exception as e:
    logger.error("Failed to instrument %s: %s", instrumentor_name, e)
```

So `openlit.init()` returns normally and the error only appears in logs at ERROR level. And because the first wrap target raises, none of the remaining three are attempted either.

Meanwhile `_instruments = ("together >= 1.3.5",)` declares 2.x supported, so the dependency check passes and openlit reports the integration as active.

### How to reproduce it?

```bash
pip install "together==2.33.2" openlit
```

```python
import logging
logging.basicConfig(level=logging.ERROR, format="LOG %(levelname)s: %(message)s")

import openlit
openlit.init(disable_metrics=True, disable_batch=True)

import together.resources.chat.completions as c
print(type(c.CompletionsResource.create).__name__)
```

Output:

```
LOG ERROR: Failed to instrument together: unable to resolve attribute 'ChatCompletions'
in path 'ChatCompletions.create' on <module 'together.resources.chat.completions'>
function
```

`function` rather than `FunctionWrapper` confirms the real class is never wrapped. Making a Together chat or image call afterwards produces no span.

### Why CI does not catch it

`sdk/python/tests/requirements.txt` lists `together` unpinned, so CI resolves 2.33.2 today — the version where none of this binds. `tests/test_together.py` then asserts only on the API response shape:

```python
assert response.model == "meta-llama/Llama-3.3-70B-Instruct-Turbo"
assert response.model == "black-forest-labs/FLUX.1-dev"
```

There are no span, tracer or exporter assertions in that file at all, so the tests pass identically whether the instrumentation attached or not.

### Anything else?

`_instruments` declares `together >= 1.3.5`, so a fix should keep working on 1.x rather than simply renaming the targets. Resolving whichever class name is present at instrument time, and skipping cleanly when neither is, would cover both and stop the next rename from taking the integration down.

Worth adding span assertions to `test_together.py` at the same time, otherwise the suite stays green through exactly this failure.

Happy to open a PR for this if useful.

basil-k-aji-dev:
Fixes #1565.

## The problem

On `together` 2.x, the Together AI instrumentation binds to nothing — `openlit.init()` returns normally and no Together span is ever produced.

`together` 2.0 renamed every resource class the instrumentor wraps:

| wrapped before | actual in `together` 2.33.2 |
| --- | --- |
| `ChatCompletions.create` | `CompletionsResource.create` |
| `AsyncChatCompletions.create` | `AsyncCompletionsResource.create` |
| `Images.generate` | `ImagesResource.generate` |
| `AsyncImages.generate` | `AsyncImagesResource.generate` |

`wrap_function_wrapper` resolves its target eagerly, so the stale names raise:

```
PathResolutionError: unable to resolve attribute 'ChatCompletions' in path
'ChatCompletions.create' on <module 'together.resources.chat.completions'>
```

`instrument_if_available` catches that and turns it into `logger.error(...)`, so nothing surfaces unless you are watching ERROR logs. And because the very first wrap target raises, the remaining three are never attempted.

## The fix

`_instruments` declares `together >= 1.3.5`, so renaming the targets would just move the breakage to 1.x users. Instead `_wrap_first_present()` resolves whichever class name is present at instrument time, 2.x name first, 1.x name second:

```python
CHAT_CLASSES = ("CompletionsResource", "ChatCompletions")
ASYNC_CHAT_CLASSES = ("AsyncCompletionsResource", "AsyncChatCompletions")
IMAGE_CLASSES = ("ImagesResource", "Images")
ASYNC_IMAGE_CLASSES = ("AsyncImagesResource", "AsyncImages")
```

That keeps both layouts instrumented, and makes any future rename degrade to one skipped target rather than aborting `_instrument` entirely.

This mirrors the approach already used by the Qdrant instrumentor in this SDK, which resolves method availability at instrument time instead of assuming a fixed surface.

## Tests

`tests/requirements.txt` pins nothing, so CI resolves `together` 2.x — the broken version — while `tests/test_together.py` asserts only on response shape:

```python
assert response.model == "meta-llama/Llama-3.3-70B-Instruct-Turbo"
```

That passes identically whether the instrumentation attached or not, which is why this went unnoticed.

`tests/test_together_instrumentation.py` asserts the binding itself. It needs no API key and makes no network calls, so it runs in CI as-is. Against the old hardcoded names all four targets fail:

```
FAIL  ChatCompletions.create        <- none of the names exist; target goes untraced
FAIL  AsyncChatCompletions.create   <- none of the names exist; target goes untraced
FAIL  Images.generate               <- none of the names exist; target goes untraced
FAIL  AsyncImages.generate          <- none of the names exist; target goes untraced
```

After the fix, all four resolve and are wrapped:

```
CompletionsResource.create         BoundFunctionWrapper   WRAPPED
AsyncCompletionsResource.create    BoundFunctionWrapper   WRAPPED
ImagesResource.generate            BoundFunctionWrapper   WRAPPED
AsyncImagesResource.generate       BoundFunctionWrapper   WRAPPED
```

`11 passed`, verified against `together` 2.33.2 on Python 3.13. `ruff check` and `ruff format --check` are clean on both changed files. (`__init__.py` has two pre-existing import-sort findings on `main`; left alone to keep this diff focused.)

## Note

The existing `test_together.py` is unchanged — it needs a live `TOGETHER_API_KEY`, so I did not want to alter its behaviour. Adding span assertions there would be a reasonable follow-up if you would like the live path covered too.

## Summary by Sourcery

Support Together AI resource-class renames while preserving instrumentation for older supported versions.

Bug Fixes:
- Restore Together AI instrumentation across both the 1.x and 2.x resource-class layouts so chat and image operations produce spans.

Enhancements:
- Resolve supported Together resource classes at instrumentation time and skip unavailable targets without aborting setup.

Tests:
- Add API-key-free regression tests that verify all Together instrumentation targets resolve and are wrapped.


basil-k-aji-dev:
Thanks — both findings were correct and are fixed in e40e6f5.

**1. Candidate selection ignored the method.** `_wrap_first_present` picked on class name alone, so a release exposing both names during a partial rename (with only one implementing the method) would resolve to the wrong class and raise the very `PathResolutionError` the helper exists to avoid, aborting the remaining targets. It now requires `hasattr(candidate, method)` and falls through to the next candidate. Added a test that registers a stub class without `create` ahead of the real candidates and asserts the real one is still chosen.

**2. The tests never exercised `_instrument`.** Correct, and this was the more useful catch — they validated the constants only, so an `_instrument` reverted to stale hardcoded targets would have passed every one of them while Together stayed untraced. Added `test_instrument_binds_every_target`, which runs `_instrument()` with a patched `wrap_function_wrapper`, asserts exactly four targets are recorded across both modules, and checks each recorded path actually resolves on the installed `together`. Simulated against the old hardcoded names it fails on 4/4:

```
FAIL  ChatCompletions.create        <- does not exist on together.resources.chat.completions
FAIL  AsyncChatCompletions.create   <- does not exist on together.resources.chat.completions
FAIL  Images.generate               <- does not exist on together.resources.images
FAIL  AsyncImages.generate          <- does not exist on together.resources.images
```

13 passed against `together` 2.33.2 on Python 3.13. `ruff check` and `ruff format --check` clean on the test file; the two `I001` findings in `__init__.py` are pre-existing on `main` and left alone to keep the diff focused.

Issue #1597, opened 2026-09-17
Mistral instrumentation fails on reasoning-model responses (chunked content), and crashes the caller when streaming


HarianthK:
**Where**: `sdk/python/src/openlit/instrumentation/mistral/utils.py`

Mistral's reasoning models (`magistral-*`) return the assistant message content as a list of chunks instead of a string: a `thinking` chunk (a list of text chunks) followed by a `text` chunk with the answer. In the SDK, `AssistantMessageContent = Union[str, List[ContentChunk]]`, and `response_as_dict` turns the chunks into a list of dicts.

Two things break on that shape:

1. **`chat.complete`**: `process_chat_response` does `" ".join(choice["message"]["content"] or "" ...)`, which raises `TypeError: sequence item 0: expected str instance, list found`. The wrapper's `except` calls `handle_exception`, so the span is exported with `ERROR` status and no output message, although the request succeeded.
2. **`chat.stream`**: `process_chunk` does `scope._llmresponse += content` on a chunk-list delta, raising `TypeError: can only concatenate str (not "list") to str`. That happens inside `TracedSyncStream.__next__`, which only catches `StopIteration`, so the error propagates into the user's `for chunk in stream:` loop and breaks the application, not just the telemetry.

Both reproduce with a response or stream built from the mistralai SDK's own models, no API key needed; I have a test that drives the real `complete` and `stream` wrappers with them.

**Expected**: the text chunk is recorded as the answer, the thinking text is recorded under `gen_ai.content.reasoning` (the attribute the OpenAI instrumentation already uses for `reasoning_content`), and the stream keeps yielding to the caller.

I have the fix ready and will open a PR referencing this issue.

PR #1598, opened 2026-09-17
fix(python-sdk): handle chunked content from Mistral reasoning models

HarianthK:
**Issue number**: #1597

### Change description:

Mistral's reasoning models (`magistral-*`) return the assistant content as a list of chunks, a `thinking` chunk followed by a `text` chunk, instead of a string. The Mistral instrumentation joined that list as if it were text: `process_chat_response` raised and the span was exported as `ERROR` with no output message, and in `process_chunk` the `TypeError` escaped `TracedSyncStream.__next__` (which only catches `StopIteration`) into the caller's `for chunk in stream:` loop.

This adds `split_content_chunks`, which returns the text and the reasoning for either shape, and uses it on both paths. The text is recorded as the answer exactly as before; the thinking text is recorded under `gen_ai.content.reasoning`, the attribute the OpenAI instrumentation already uses for `reasoning_content`; and the stream keeps yielding every chunk. String content is unaffected.

Two tests in `tests/test_mistral_chunked_content.py` drive the real `complete` and `stream` wrapper factories with a response and a stream built from the mistralai SDK's own models (the same approach as `test_anthropic_stream_span_lifecycle.py`), so they need no API key. On `main` the first fails with the span marked as failed and the second with the `TypeError`; both pass with the change. `pylint` rates the changed files 10/10.

### Checklist

* [ ] Changes are documented

## Summary by Sourcery

Support chunked assistant content from Mistral reasoning models while preserving streamed output and separating reasoning from the final answer.

Bug Fixes:
- Handle Mistral reasoning model responses whose assistant content is represented as thinking and text chunks.
- Prevent chunked reasoning content from failing synchronous streaming and ensure all stream chunks remain available to callers.

Enhancements:
- Record reasoning text separately under the GenAI reasoning content attribute while preserving the answer as the output message.
- Support both string and chunked assistant content formats without changing existing string-content behavior.

Tests:
- Add regression coverage for complete and stream wrappers using Mistral SDK response models without requiring an API key.


HarianthK:
On the reasoning-content point for whoever reviews: the new attribute is set only inside the existing `capture_message_content` gate, the same switch that already controls whether prompts and completions are recorded, so with content capture off nothing new is exported. It is also the same `gen_ai.content.reasoning` attribute the OpenAI instrumentation already sets for `reasoning_content` under that gate.

PR #1608, opened 2026-09-18
fix(bedrock): end converse_stream span on early close

RichardoMrMu:
> [!IMPORTANT]
> 2. PR title follows Conventional Commit format, for example `feat: ...` or
>    `fix(client): ...`.

**Issue number**: #1545

### Change description:

`TracedSyncStream` in the Bedrock instrumentation finalized its span only in the
`StopIteration` branch of `__next__`. `__exit__` just forwarded to the wrapped
stream and never touched the span, and the class had no `close()`, so
`__getattr__` forwarded `close` to the boto3 response dict and raised
`AttributeError`. Any caller that stopped consuming a `converse_stream` response
early left the span recording forever, so it was never exported and its
telemetry was lost.

This applies the same shape as the merged Anthropic fix (#1461):

- `_finalize_streaming_span()` completes and ends the span exactly once, guarded
  by `_streaming_response_processed`.
- `__next__` calls it from the `StopIteration` branch as before.
- `__exit__` wraps the delegation in `try/finally` and finalizes on every
  non-exception exit, so a `break` inside the `with` block no longer leaks.
- `close()` is added, because unlike the Anthropic wrapper this one wraps the
  boto3 response *dict*, which has no `close` of its own.

Bedrock has no async `converse_stream` path, so there is no async twin to change.

### Verification

`sdk/python/tests/test_bedrock_stream_span_lifecycle.py` drives the real
`converse_stream` wrapper factory with a synthetic boto3-shaped client and event
stream, asserting the span ends exactly once on each exit path. Before the change
three of the four cases fail (two export no span at all; `close()` raises
`AttributeError`); after it all four pass.

```
pytest tests/ -k stream_span_lifecycle --asyncio-mode=auto
18 passed
```

`pylint` on the changed implementation reports the same set of pre-existing
messages as `main` with no new ones, and the new test module is clean.

## Summary by Sourcery

Prevent Bedrock converse_stream telemetry from leaking by finalizing streaming spans on every termination path.

Bug Fixes:
- Ensure Bedrock converse_stream spans are finalized and exported when consumers stop early, exit a context block, encounter an exception, or explicitly close the stream.

Enhancements:
- Make streaming span finalization idempotent so exhaustion, close, and context-manager cleanup cannot end the span more than once.

Tests:
- Add regression coverage for full consumption, early loop breaks, early context-manager exits, explicit close, and exceptions during Bedrock streaming.


RichardoMrMu:
I pushed one more commit to this branch after re-reading the earlier attempt at
this fix in #1546 and its automated review.

That review pointed out that finalizing only on a clean exit still leaks the span
when an exception escapes the `with` block, and it was right — I reproduced it
against my own branch: raising inside the block left `get_finished_spans()` empty.

So `__exit__` now finalizes on every exit rather than only the clean one, and when
an exception escaped it records that exception on the span first, so the span
carries the error instead of looking clean. The suite gains a fifth case covering
that path.

Current state of the suite on this branch:

```
pytest tests/test_bedrock_stream_span_lifecycle.py --asyncio-mode=auto
5 passed

pytest tests/ -k stream_span_lifecycle --asyncio-mode=auto
19 passed
```

`pylint` on the changed implementation still reports only the messages that are
already present on `main`, with no new ones.

For attribution: #1546 by @wangzhengzhuo05 proposed the same approach eight days
earlier and was self-closed as stale this morning, before I opened this PR. I
found it only afterwards, through the cross-reference on #1545. Happy to close
this in favour of that branch if maintainers would rather take the original.

PR #1610, opened 2026-09-19
fix(premai): make the streaming wrapper iterable and end its span once

### LINKED ISSUE #1600: Bug: premai streaming raises TypeError, the traced wrapper is not an iterator
### Component

OpenLIT Python SDK

### What happened?

Installing openlit turns a working premai stream into a crash. On `cc00e01`, `TracedSyncStream` in `sdk/python/src/openlit/instrumentation/premai/premai.py` defines `__iter__` at line 85 and no `__next__` anywhere in the file. Line 87 pulls one chunk, 88 processes it, and 89 returns that chunk rather than an iterator.

```
TypeError: iter() returned non-iterator of type 'dict'
```

`next(stream)` fails too, because CPython looks `__next__` up on the type and `__getattr__` cannot serve it.

This is not another instance of the early-close class in #1545, #1549, #1553, #1555, #1559, #1561 and #1593. Those spans end late; this stream cannot be iterated at all. A scan of every file under `sdk/python/src/openlit/instrumentation/` at `cc00e01` finds premai as the only one with `def __iter__` and no `def __next__`.

The `finally` at line 91 is the second half. It ends the span after that single chunk, so the one exported span carries the first delta only, and `gen_ai.response.id` stays empty because `utils.py:300-302` reads the id inside the finish-reason branch, which an aborted iteration never reaches.

### Steps to reproduce

Save as `repro.py` and run from `sdk/python` with `PYTHONPATH=src`. No premai account and no premai install are needed: `process_chunk` goes through `response_as_dict`, so plain dicts suffice, and the mistral and openai stream tests here already do that.

```python
from openlit.__helpers import OpenlitConfig
from openlit.instrumentation.premai.premai import chat
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter

OpenlitConfig.reset_to_defaults()
exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))

CHUNKS = [
    {"id": "", "choices": [{"delta": {"content": "Hello"}, "finish_reason": None}]},
    {"id": "cmpl-1",
     "choices": [{"delta": {"content": " world"}, "finish_reason": "stop"}],
     "usage": {"prompt_tokens": 5, "completion_tokens": 3, "total_tokens": 8}},
]

wrapper = chat(
    version="1.0.0", environment="test", application_name="repro",
    tracer=provider.get_tracer(__name__), pricing_info={},
    capture_message_content=True, metrics=None, disable_metrics=True,
    event_provider=None,
)

stream = wrapper(lambda *a, **k: iter(CHUNKS), None, (), {"model": "gpt-4o-mini", "stream": True})
try:
    list(stream)
except TypeError as exc:
    print(f"TypeError: {exc}")

for s in exporter.get_finished_spans():
    print("output tokens:", s.attributes.get("gen_ai.usage.output_tokens"))
    print("response id:  ", repr(s.attributes.get("gen_ai.response.id")))
```

Output on `cc00e01`:

```
TypeError: iter() returned non-iterator of type 'dict'
output tokens: 3
response id:   ''
```

### Expected behavior

`list(stream)` returns both chunks, and exactly one span is exported when the stream is exhausted, carrying the content of every chunk and the response id from the last one.

`openai/openai.py` is the model: `__iter__` returns `self` at line 92, and `__next__` finalizes on `StopIteration` at line 126. groq and mistral do the same.

### Environment

- OpenLIT Python SDK: `cc00e01` (1.45.0)
- Python: 3.12.3
- premai: not installed, and not needed for the reproduction. The instrumentation activates on `premai >= 0.3.79`.

### Duplicate search

Run with `--state all`, because both default to open and return nothing:

- `gh issue list --repo openlit/openlit --search "premai" --state all` returns #675, #758, #1438 and #1460. None is about iteration.
- `gh issue list --repo openlit/openlit --search "iterator stream" --state all` returns #391, a 2024 report about a generator used as a context manager, and #1549, which is Cohere early-close.
- `gh pr list --repo openlit/openlit --search "premai" --state all` returns #1464, #1018, #774, #770, #686, #629, #628, #627, #616 and #519, all merged.

#1464 deserves naming, because it touched this exact class. It closed #1460 and edited `premai.py`, replacing the `with tracer.start_as_current_span(...)` in the finalization block with `with self._span`. It left `__iter__` as it was.

### Not checked

No live premai call. The reproduction drives the wrapper with dicts, so it claims the iterator protocol and the span contents that follow from it, not premai's wire format.

One related defect stays out of this report. `__exit__` is a bare `pass` and there is no `close()`, so `with stream as s: break` never finalizes the span. That is the early-close class above, and it needs a separate decision about what premai's stream object supports.

### Are you willing to submit a PR?

Yes. A branch with the fix and three regression tests is ready, including one that pins the double-finalize guard. All three fail against a pristine `origin/main` tree.

basil-k-aji-dev:
Fixes #1600.

## The two faults

`TracedSyncStream` in `premai.py` defined `__iter__` with a `__next__` body — it pulled one chunk, ran `process_chunk`, and returned that chunk where an iterator belongs. `for chunk in stream:` therefore raised `TypeError: iter() returned non-iterator`, so installing openlit turned a working premai stream into a crash. A scan of every file under `src/openlit/instrumentation/` confirms premai was the only one with `__iter__` and no `__next__`.

The `finally` wrapped around it was the second fault. It finalized the span after the first chunk, so the single exported span carried the first delta only and `gen_ai.response.id` stayed empty — the id is read in the finish-reason branch, which an aborted iteration never reaches.

## The fix

Split them the way `groq/groq.py` already does: `__iter__` returns `self`, `__next__` finalizes on `StopIteration`, and `_finalize_streaming_span()` behind a `_streaming_response_processed` flag runs from `close()` and `__exit__` too, so a caller that breaks early exports the span instead of leaking it. `__enter__`/`__exit__` now forward to the wrapped stream as groq's do.

## Tests

`tests/test_premai_stream_span_lifecycle.py`, six tests driving the real wrapper factory with a premai-shaped stream: iteration works at all, full consumption exports one span carrying both deltas and the response id, an early `break` still exports, `close()` twice finalizes once, context-manager exit finalizes, and the context token is detached.

**All six fail against unpatched `src/`** — the first with the `TypeError` from the issue.

Locally: `pylint` 10.00/10 on the changed files; full suite `687 passed` against `681` on `main`, with the same 4 pre-existing failures (`test_anthropic_stream_span_lifecycle` ×2, `test_google_adk`, `test_litellm_reproduction`) and the same 11 collection errors from optional SDKs I do not have installed. No new failure.

---

**On duplication, stated plainly.** @chrikrah wrote in this issue that a branch was ready, and @RichardoMrMu offered to stay out of all four while waiting for it. This PR is not an attempt to step over either of them. The analysis in the issue is @chrikrah's, it is what I worked from, and if their branch lands this one should be closed without ceremony — @chrikrah, say the word and I will close it myself. I have credited the analysis in the commit body rather than presenting it as my own finding.

## Summary by Sourcery

Fix Premai stream iteration and reliably finalize its tracing span across every stream lifecycle.

Bug Fixes:
- Fix the Premai streaming wrapper so it is properly iterable and preserves all streamed response data.
- Ensure streaming spans are finalized exactly once on exhaustion, early exit, close, or context-manager exit.

Enhancements:
- Forward stream context-manager behavior and exception suppression to the wrapped stream.

Tests:
- Add lifecycle regression coverage for iteration, complete and interrupted consumption, repeated close, context-manager exit, exception suppression, response attributes, and context detachment.


basil-k-aji-dev:
Thanks — that one is right, and it is mine rather than inherited.

`__exit__` was `pass` on `main`; this PR introduced the forwarding to the wrapped stream, so dropping its return value is a regression this PR would have shipped. A wrapped context manager that suppresses an exception says so with a truthy `__exit__`, and returning `None` turned that answer into a re-raise.

Fixed in 8b232ec, with a test using a stream whose `__exit__` returns `True`. It raises `ValueError` out of the `with` block when the value is dropped, so it fails against the version you reviewed.

Worth noting for whoever picks it up: `groq/groq.py` and `mistral/mistral.py` have the same gap, and they were the pattern I copied. I have left them alone to keep this PR to one issue, but they are a reasonable follow-up.

PR #1612, opened 2026-09-19
fix(browser_use/crawl4ai): record operation duration in seconds

### LINKED ISSUE #1602: Bug: browser_use and crawl4ai set gen_ai.client.operation.duration in milliseconds, every other instrumentation uses seconds
### Component

OpenLIT Python SDK

### What happened?

Nine sites in `browser_use` and `crawl4ai` compute `(end_time - start_time) * 1000` and set the result as the `gen_ai.client.operation.duration` span attribute. Every other site in the SDK that writes that name writes seconds.

A two-second crawl reports `2000.0`. A two-second model call reports `2.0`. Both land under one attribute name, so any average or percentile over it mixes the two by 1000.

| file | lines |
|---|---|
| `sdk/python/src/openlit/instrumentation/browser_use/browser_use.py` | 76, 104 |
| `sdk/python/src/openlit/instrumentation/browser_use/async_browser_use.py` | 93, 121 |
| `sdk/python/src/openlit/instrumentation/crawl4ai/crawl4ai.py` | 120, 183 |
| `sdk/python/src/openlit/instrumentation/crawl4ai/async_crawl4ai.py` | 151, 194, 317 |

These are span attributes, not the histogram. Neither instrumentation calls `.record()` on any instrument:

```
$ git grep -n '\.record(' origin/main -- 'sdk/python/src/openlit/instrumentation/browser_use/*' 'sdk/python/src/openlit/instrumentation/crawl4ai/*'
$
```

The histogram's advisory buckets therefore never see these values. The project's own normative table decides the unit, at `agent-guides/js-sdk-genai-instrumentation.md:224`:

```
| `gen_ai.client.operation.duration` | `s` | [0.01, 0.02, ... 81.92] | Always |
```

The naming convention in the code agrees. Where this SDK means milliseconds it says so in the name: `gen_ai.operation.duration_ms` in `llamaindex/utils.py:229`, `gen_ai.agent.duration_ms` in `claude_agent_sdk/utils.py:575`. `otel/metrics.py:243` creates the matching histogram and `:246` sets `unit="s"` on it.

### Steps to reproduce

Save as `repro.py` and run from `sdk/python` with `PYTHONPATH=src`. Neither `crawl4ai` nor `browser_use` needs an install. The script drives the wrapper directly with a stub and patches the clock, so the elapsed time is exactly 2.5 seconds.

```python
from unittest.mock import patch
from openlit.__helpers import OpenlitConfig
from openlit.instrumentation.crawl4ai import crawl4ai as c4
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter

OpenlitConfig.reset_to_defaults()
exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))

wrapper = c4.general_wrap(
    gen_ai_endpoint="crawl", version="1.0.0", environment="test",
    application_name="repro", tracer=provider.get_tracer(__name__),
    pricing_info={}, capture_message_content=True, metrics={}, disable_metrics=True,
)

with patch.object(c4.time, "time", side_effect=[100.0, 102.5]):
    wrapper(lambda *a, **k: None, None, (), {"url": "https://example.com"})

for s in exporter.get_finished_spans():
    print(s.name, "->", s.attributes.get("gen_ai.client.operation.duration"))
```

Output on `cc00e01`:

```
crawl https://example.com -> 2500.0
```

### Expected behavior

A 2.5 second operation sets `gen_ai.client.operation.duration` to `2.5`, matching what a 2.5 second model call sets, so one query can aggregate the attribute across instrumentations.

### Environment

- OpenLIT Python SDK: `cc00e01` (1.45.0)
- Python: 3.12.3
- `crawl4ai` and `browser_use`: not installed, and not needed for the reproduction.

### Additional context

**Duplicate search**

Run with `--state all`; both default to open and return nothing:

- `gh issue list --repo openlit/openlit --search "crawl4ai" --state all` returns #758 and #544, both closed and unrelated.
- `gh issue list --repo openlit/openlit --search "browser_use" --state all` returns #1242, closed and unrelated.
- `gh pr list --repo openlit/openlit --search "crawl4ai duration" --state all` returns nothing.

None of the 26 open pull requests touches these four files.

**Who this affects**

`src/client` never reads this attribute. `git grep -c 'gen_ai.client.operation.duration' origin/main -- 'src/client/*'` returns zero, and the UI reads latency from the ClickHouse `Duration` column instead. No OpenLIT dashboard depends on the current values, so correcting them breaks nothing. It affects anyone querying `SpanAttributes` directly, where the two units look alike.

**Not checked**

No live `crawl4ai` or `browser_use` call. The reproduction drives the wrapper with a stub and a patched clock.

One of the nine sites, in `async_crawl4ai._handle_async_generator`, sets its attribute after the span has closed, so the SDK drops it with `Setting attribute on ended span`. That is a separate defect and stays out of this report. It means eight of the nine values reach a backend today.

### Pre-submission checklist

Yes. A branch dropping the factor at all nine sites, with a regression test that drives the real wrapper code, is ready.

basil-k-aji-dev:
Fixes #1602.

## The fault

Nine sites in `browser_use` and `crawl4ai` computed `(end_time - start_time) * 1000` and set the result as `gen_ai.client.operation.duration`:

```
browser_use/browser_use.py:76,104
browser_use/async_browser_use.py:93,121
crawl4ai/crawl4ai.py:120,183
crawl4ai/async_crawl4ai.py:151,194,317
```

The attribute is seconds everywhere else, and the repository says so in three independent places:

- `otel/metrics.py:243-248` creates the histogram with `unit="s"` and advisory buckets ending at **81.92**
- `mem0/mem0.py:86` and `mem0/async_mem0.py:78` set the same attribute from an unscaled duration
- the normative table at `agent-guides/js-sdk-genai-instrumentation.md:224` lists the unit as `s`

So a two-and-a-half second crawl arrived as `2500`, past the last bucket. Every browser_use and crawl4ai operation landed in the overflow bucket, and the latency histogram carried no usable shape for either integration.

## The fix

Drop the factor at all nine sites, and rename `duration_ms` → `duration_s` (20 identifiers) so the name cannot go on claiming milliseconds after the value stopped being milliseconds.

## Tests

`tests/test_operation_duration_unit.py`:

- Two tests drive the **real** `general_wrap` factories for crawl4ai and browser_use with a patched clock advancing exactly 2.5s, and assert the recorded attribute is `2.5`, not `2500`. They measure what the instrumentation records rather than re-deriving the arithmetic.
- A third scans all four source files for `* 1000` and `duration_ms`, so a future edit cannot quietly reintroduce one of the nine. Its failure output names every offending line.

**All three fail against unpatched `src/`.**

Locally: `pylint` 10.00/10; full suite `684 passed` against `681` on `main`, same 4 pre-existing failures and 11 collection errors. No new failure.

---

**On duplication, stated plainly.** @chrikrah wrote in this issue that a branch was ready, and @RichardoMrMu offered to stay out of all four while waiting for it. This PR is not an attempt to step over either of them. The analysis in the issue is @chrikrah's, it is what I worked from, and if their branch lands this one should be closed without ceremony — @chrikrah, say the word and I will close it myself. I have credited the analysis in the commit body rather than presenting it as my own finding.

## Summary by Sourcery

Correct operation duration units for browser_use and crawl4ai instrumentation and add safeguards against future regressions.

Bug Fixes:
- Record browser_use and crawl4ai operation durations in seconds, restoring correctly scaled latency telemetry and metric histograms.

Enhancements:
- Use seconds-based duration naming consistently across synchronous, asynchronous, error, and streaming operation paths.

Tests:
- Add regression coverage using real browser_use and crawl4ai wrappers, plus source checks that prevent millisecond scaling from being reintroduced.


basil-k-aji-dev:
Fair, and the criticism lands: a guard that matches one spelling is not much of a guard. `1000.0`, `1_000`, different spacing and `1000 * (end - start)` are the same bug, and all four would have walked past it.

Rewritten in 18865f5 to judge shape rather than text. It walks the AST and flags a multiplication where one side is a number of at least a thousand and the other mentions a start time, so the spelling stops mattering.

Added a second test that exercises the guard itself — five ways of writing the bug must be caught, three innocent expressions must not:

```python
"d = (end_time - start_time) * 1000"        -> caught
"d = (end_time - start_time)*1000.0"        -> caught
"d = (end_time - start_time) * 1_000"       -> caught
"d = 1000 * (end_time - start_time)"        -> caught
"d = (end_time - self.start_time) * 1000"   -> caught

"d = end_time - start_time"                 -> clean
"d = tokens * 1000"                         -> clean
"d = (end_time - start_time) * 1"           -> clean
```

Without that, a future edit could weaken the guard and nothing would notice. All three original tests still fail against `main`'s source.

